PR #47, opened 2018-02-18
Resource counting and limiting

### LINKED ISSUE #34: Limit result size
Limit result list(s) length (overall items count or items count for each list) or limit result size in bytes.

Totktonada:
Fixes #34.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2,7 +2,8 @@ std = {
     read_globals = {'box', 'require', 'debug', 'pcall', 'xpcall', 'tostring',
         'tonumber', 'type', 'assert', 'ipairs', 'math', 'error', 'string',
         'table', 'pairs', 'os', 'select', 'unpack', 'dofile', 'next',
-        'getmetatable', 'setmetatable', 'rawget', 'print'
+        'getmetatable', 'setmetatable', 'rawget', 'print', 'shard_status',
+	'loadstring'
     },
     globals = {'package'}
 }
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ default:
 .PHONY: lint
 lint:
 	luacheck graphql/*.lua test/local/*.lua test/testdata/*.lua \
-		--no-redefined --no-unused-args
+		test/common/lua/*.lua --no-redefined --no-unused-args
 
 .PHONY: test
 test: lint
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ default:
 .PHONY: lint
 lint:
 	luacheck graphql/*.lua test/local/*.lua test/testdata/*.lua \
-		test/common/lua/*.lua --no-redefined --no-unused-args
+		test/common/*test.lua test/common/lua/*.lua \
+		--no-redefined --no-unused-args
 
 .PHONY: test
 test: lint
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,10 @@ test: lint
 		pip install tarantool && \
 		cd test && ./test-run.py
 
+.PHONY: pure-test
+pure-test:
+	cd test && ./test-run.py
+
 .PHONY: clean
 clean:
 	rm -rf test/var
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -9,6 +9,8 @@ local avro_schema = require('avro_schema')
 local utils = require('graphql.utils')
 
 local accessor_general = {}
+local RESULTING_OBJECT_CNT_MAX = 10000
+local FETCHED_OBJECT_CNT_MAX = 10000
 
 --- Validate and compile set of avro schemas (with respect to service fields).
 --- @tparam table schemas map where keys are string names and values are
@@ -550,7 +552,8 @@ end
 ---
 --- * `count` (number),
 --- * `objs` (table, list of objects),
---- * `pivot_found` (boolean).
+--- * `pivot_found` (boolean),
+--- * `statistics` (table, per-query statistics).
 ---
 --- @tparam cdata tuple flatten representation of an object to process
 ---
@@ -562,7 +565,9 @@ end
 --- * `do_filter` (boolean, whether we need to filter out non-matching
 ---   objects),
 --- * `pivot_filter` (table, set of fields to match the objected pointed by
----   `offset` arqument of the GraphQL query)
+---   `offset` arqument of the GraphQL query),
+--- * `resulting_object_cnt_max` (number),
+--- * `fetched_object_cnt_max` (number).
 ---
 --- @return nil
 ---
@@ -580,6 +585,14 @@ local function process_tuple(state, tuple, opts)
     local filter = opts.filter
     local do_filter = opts.do_filter
     local pivot_filter = opts.pivot_filter
+    local qstats = state.statistics
+    local resulting_object_cnt_max = opts.resulting_object_cnt_max
+    local fetched_object_cnt_max = opts.fetched_object_cnt_max
+    qstats.fetched_object_cnt = qstats.fetched_object_cnt + 1
+    assert(qstats.fetched_object_cnt <= fetched_object_cnt_max,
+            ('fetched object count[%d] exceeds limit[%s] '..
+                    '(set `fetched_object_cnt_max` in accessor)'):format(
+                    qstats.fetched_object_cnt, fetched_object_cnt_max))
 
     -- skip all items before pivot (the item pointed by offset)
     local obj = nil
@@ -604,10 +617,12 @@ local function process_tuple(state, tuple, opts)
     -- add the matching object, update count and check limit
     state.objs[#state.objs + 1] = obj
     state.count = state.count + 1
+    qstats.resulting_object_cnt = qstats.resulting_object_cnt + 1
+    assert(qstats.resulting_object_cnt <= resulting_object_cnt_max,
+            ('returning object count[%d] exceeds limit[%s] '..
+                    '(set `resulting_object_cnt_max` in accessor)'):format(
+                    qstats.resulting_object_cnt, resulting_object_cnt_max))
     if limit ~= nil and state.count >= limit then
-        assert(limit == nil or state.count <= limit,
-            ('count[%d] exceeds limit[%s] (in for)'):format(
-            state.count, limit))
         return false
     end
     return true
@@ -631,8 +646,11 @@ end
 --- @tparam table args table of arguments passed within the query except ones
 --- that forms the `filter` parameter
 ---
+--- @tparam table extra table which contains extra information related to
+--- current select and the whole query
+---
 --- @treturn table list of matching objects
-local function select_internal(self, collection_name, from, filter, args)
+local function select_internal(self, collection_name, from, filter, args, extra)
     assert(type(self) == 'table',
         'self must be a table, got ' .. type(self))
     assert(type(collection_name) == 'string',
@@ -694,6 +712,7 @@ local function select_internal(self, collection_name, from, filter, args)
         count = 0,
         objs = {},
         pivot_found = false,
+        statistics = extra.qcontext.statistics
     }
 
     -- read only process_tuple options
@@ -703,6 +722,8 @@ local function select_internal(self, collection_name, from, filter, args)
         filter = filter,
         do_filter = not full_match,
         pivot_filter = nil, -- filled later if needed
+        resulting_object_cnt_max = self.settings.resulting_object_cnt_max,
+        fetched_object_cnt_max = self.settings.fetched_object_cnt_max
     }
 
     if index == nil then
@@ -791,7 +812,10 @@ end
 ---
 --- @tparam table opts `schemas`, `collections`, `service_fields` and `indexes`
 --- to give the data accessor all needed meta-information re data; the format is
---- shown below
+--- shown below; additional attributes `resulting_object_cnt_max` and
+--- `fetched_object_cnt_max` are optional natural numbers whish help to control
+--- query behaviour in case it requires more resources than expected _(default
+--- value is 10.000)_
 ---
 --- @tparam table funcs set of functions (`is_collection_exists`, `get_index`,
 --- `get_primary_index`) allows this abstract data accessor behaves in the
@@ -837,6 +861,10 @@ function accessor_general.new(opts, funcs)
     local collections = opts.collections
     local service_fields = opts.service_fields
     local indexes = opts.indexes
+    local resulting_object_cnt_max = opts.resulting_object_cnt_max or
+                                     RESULTING_OBJECT_CNT_MAX
+    local fetched_object_cnt_max = opts.fetched_object_cnt_max or
+                                   FETCHED_OBJECT_CNT_MAX
 
     assert(type(schemas) == 'table',
         'schemas must be a table, got ' .. type(schemas))
@@ -846,6 +874,12 @@ function accessor_general.new(opts, funcs)
         'service_fields must be a table, got ' .. type(service_fields))
     assert(type(indexes) == 'table',
         'indexes must be a table, got ' .. type(indexes))
+    assert(type(resulting_object_cnt_max) == 'number' and
+                resulting_object_cnt_max > 0,
+        'resulting_object_cnt_max must be natural number')
+    assert(type(fetched_object_cnt_max) == 'number' and
+                fetched_object_cnt_max > 0,
+        'fetched_object_cnt_max must be natural number')
 
     local models = compile_schemas(schemas, service_fields)
     validate_collections(collections, schemas, indexes)
@@ -861,10 +895,14 @@ function accessor_general.new(opts, funcs)
         models = models,
         index_cache = index_cache,
         funcs = funcs,
+        settings = {
+            resulting_object_cnt_max = resulting_object_cnt_max,
+            fetched_object_cnt_max = fetched_object_cnt_max
+        }
     }, {
         __index = {
             select = function(self, parent, collection_name, from,
-                    filter, args)
+                    filter, args, extra)
                 assert(type(parent) == 'table',
                     'parent must be a table, got ' .. type(parent))
                 assert(from == nil or type(from) == 'table',
@@ -877,8 +915,13 @@ function accessor_general.new(opts, funcs)
                     'from must be nil or from.connection_name ' ..
                     'must be a string, got ' ..
                     type((from or {}).connection_name))
+                -- use `extra.qcontext` to store per-query variables
+                extra.qcontext.statistics = extra.qcontext.statistics or {
+                    resulting_object_cnt = 0,
+                    fetched_object_cnt = 0
+                }
                 return select_internal(self, collection_name, from, filter,
-                    args)
+                    args, extra)
             end,
             list_args = function(self, collection_name)
                 -- get name of field of primary key
--- a/graphql/core/execute.lua
+++ b/graphql/core/execute.lua
@@ -88,7 +88,10 @@ local function buildContext(schema, tree, rootValue, variables, operationName)
     rootValue = rootValue,
     variables = variables,
     operation = nil,
-    fragmentMap = {}
+    fragmentMap = {},
+    -- The field is passed to resolve function within info attribute.
+    -- Can be used to store any data within one query.
+    qcontext = {}
   }
 
   for _, definition in ipairs(tree.definitions) do
@@ -222,7 +225,8 @@ local function getFieldEntry(objectType, object, fields, context)
     fragments = context.fragmentMap,
     rootValue = context.rootValue,
     operation = context.operation,
-    variableValues = context.variables
+    variableValues = context.variables,
+    qcontext = context.qcontext
   }
 
   local resolvedObject = (fieldType.resolve or defaultResolver)(object, arguments, info)
--- a/graphql/tarantool_graphql.lua
+++ b/graphql/tarantool_graphql.lua
@@ -286,7 +286,9 @@ gql_type = function(state, avro_schema, collection, collection_name)
                         destination_args_names = destination_args_names,
                         destination_args_values = destination_args_values,
                     }
-
+                    local extra = {
+                        qcontext = info.qcontext
+                    }
                     local object_args_instance = {} -- passed to 'filter'
                     local list_args_instance = {} -- passed to 'args'
                     for k, v in pairs(args_instance) do
@@ -302,7 +304,7 @@ gql_type = function(state, avro_schema, collection, collection_name)
                     end
                     local objs = accessor:select(parent,
                         c.destination_collection, from,
-                        object_args_instance, list_args_instance)
+                        object_args_instance, list_args_instance, extra)
                     assert(type(objs) == 'table',
                         'objs list received from an accessor ' ..
                         'must be a table, got ' .. type(objs))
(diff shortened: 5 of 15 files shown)


Makefile
```diff
@@ -6,13 +6,20 @@ lint:
 	luacheck graphql/*.lua test/local/*.lua test/testdata/*.lua \
 		--no-redefined --no-unused-args
 
-.PHONY: test
-test: lint
+.PHONY:
```
Totktonada: .PNOHY w/o target name works? Even if so, I would use the same syntax for every rule.

Makefile
```diff
@@ -6,13 +6,20 @@ lint:
...
+pure_test:
```
Totktonada: As I know, make targets usually are dash-separated words: https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets

Makefile
```diff
@@ -6,13 +6,20 @@ lint:
...
+
+.PHONY: virtualenv
+virtualenv:
 	virtualenv -p python2.7 ./.env-2.7
 	. ./.env-2.7/bin/activate && \
 		pip install -r ./test-run/requirements.txt && \
-		pip install tarantool && \
-		cd test && ./test-run.py
+		pip install tarantool
+
```
Totktonada: Two expty lines instead of one.

Makefile
```diff
@@ -6,13 +6,20 @@ lint:
...
 		--no-redefined --no-unused-args
 
-.PHONY: test
-test: lint
+.PHONY:
+pure_test:
+	cd test && ./test-run.py
+
+.PHONY: virtualenv
+virtualenv:
```
Totktonada: Do the `source` (dot) command works over make rules? My guessing is that each rule (or likely even each line) produce the new subshell. Please, check that `make test` executes tests within the virtualenv environment.

test/common/suite.ini
```diff
@@ -0,0 +1,5 @@
+[default]
+core = app
+description = tests different setups simultaneously
+lua_libs = lua/multirunner.lua lua/td_user_order.lua
+
```
Totktonada: extra newline

test/common/lua/multirunner.lua
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env tarantool
+local json = require 'json'
+local netb = require'net.box'
+local shard = require 'shard'
```
Totktonada: Proposed to use the style `require('smth')` everywhere.

graphql/accessor_general.lua
```diff
@@ -7,7 +7,7 @@
 local json = require('json')
 local avro_schema = require('avro_schema')
 local utils = require('graphql.utils')
-
+local OBJECT_CNT_MAX = 10000
```
Totktonada: It is better to move everything except requires after the module variable definition (line 11).

graphql/accessor_general.lua
```diff
@@ -631,8 +635,10 @@ end
 --- @tparam table args table of arguments passed within the query except ones
 --- that forms the `filter` parameter
 ---
+--- @tparam table info extra information related to query, passed from core
+---
 --- @treturn table list of matching objects
-local function select_internal(self, collection_name, from, filter, args)
+local function select_internal(self, collection_name, from, filter, args, info)
```
Totktonada: Why do you pass `info` here, not `qcontext`? I looks for me as an abstraction leak.

graphql/accessor_general.lua
```diff
@@ -864,7 +872,7 @@ function accessor_general.new(opts, funcs)
     }, {
         __index = {
             select = function(self, parent, collection_name, from,
-                    filter, args)
+                    filter, args, info)
```
Totktonada: Please, update related ldoc comment in the `tarantool_graphql.lua`, we have documented API for data accessors in the file.

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env tarantool
+local multirunner = require "multirunner"
+local data = require "td_user_order"
+local test_run = require('test_run').new()
+
+box.cfg({})
```
Totktonada: What is the purpose of this box.cfg?

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env tarantool
+local multirunner = require "multirunner"
+local data = require "td_user_order"
```
Totktonada: Proposed to use the same style `require('smth')` everywhere.

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,75 @@
...
+local test_run = require('test_run').new()
+
+box.cfg({})
+
+-- require in-repo version of graphql/ sources despite current working directory
+fio = require('fio')
```
Totktonada: Why do you use `local` spodarically within this file? The code will not work with require('strict').on() or debug tarantool's build? I propose to add as many as possible files into `make lint` check, including this one.

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,75 @@
...
+    print(('RESULT %s'):format(contains(result, "count[4] exceeds limit[3]")))
+
+end
+
+multirunner.run(test_run,
+                data.init_spaces_script,
+                data.drop_spaces_script,
+                run)
+os.exit()
+
```
Totktonada: Extra empty line at end of the file.

test/common/lua/multirunner.lua
```diff
@@ -0,0 +1,95 @@
...
+    }, 'shard_redundancy');
+
+    for _, shard_serv in ipairs(shard_status().online) do
+        c = netb.connect(shard_serv.uri)
+        c:eval(init_script)
+    end
+
+    callback("Shard 2x2", shard)
+
+    for _, shard_serv in ipairs(shard_status().online) do
```
Totktonada: shard_serv → shard_server, netb → netbox. There is no reason to cut two symbols and reduce the code readability.

test/common/lua/multirunner.lua
```diff
@@ -0,0 +1,95 @@
...
+    for _, name in ipairs(servers) do
+        test_run:cmd(drop_cluster_cmd3:format(name))
+    end
+end
+
+-- Run tests on multiple accessors and configurations.
+-- Feel free to add more configurations.
+local function run(test_run, init_script, cleanup_script, callback)
+
+    local SERVERS = {'shard1', 'shard2', 'shard3', 'shard4'};
```
Totktonada: There is no reason to write a local function variable in uppercase (esp. one that you change within the function), because such variable expected to be global constant.

test/common/lua/multirunner.lua
```diff
@@ -0,0 +1,95 @@
...
+            { uri = instance_uri('1'), zone = '0' },
+            { uri = instance_uri('2'), zone = '1' },
+            { uri = instance_uri('3'), zone = '2' },
+            { uri = instance_uri('4'), zone = '3' },
+        },
+        login = 'guest',
+        password = '',
+        redundancy = 2,
+        monitor = false
+    }, 'shard_redundancy');
```
Totktonada: The third argument is a suite name and will forwarded into the create_cluster function. It does not look you want to use this module from the 'shard_redundancy' suite.

test/common/lua/multirunner.lua
```diff
@@ -0,0 +1,95 @@
...
+
+    loadstring(init_script)()
+    callback("Local (box)", nil)
+    loadstring(cleanup_script)()
+end
+
+return {
+    run = run
+}
+
```
Totktonada: Extra empty line at end of the file.

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }
+    ]]
+
+    local gql_query = gql_wrapper:compile(query)
+    local variables = {
+        user_id = 5,
+    }
+    local ok, result = pcall(gql_query.execute, gql_query, variables)
```
Totktonada, line 63: Use `_` if you want to ignore result.

graphql/accessor_general.lua
```diff
@@ -877,8 +885,14 @@ function accessor_general.new(opts, funcs)
                     'from must be nil or from.connection_name ' ..
                     'must be a string, got ' ..
                     type((from or {}).connection_name))
+                -- use `info.qcontext` to store pre-query variables
+                info.qcontext.stats = info.qcontext.stats or {
+                    objects_cnt = 0,
+                    objects_fetched = 0
```
Totktonada: `object_cnt` name looks to general for me. Proposed: `resulting_objects_cnt` and `fetched_objects_cnt` and corrspondingly named limits: `resulting_object_cnt_max` and `fetched_objects_cnt_max`.

General suggestion: be strong with names. Try hard to use one name for one abstraction within the project. Use as much specific names as possible.

Totktonada (commented):
Random nitpicking. You can ignore most of these comments. Just share my feeling about some style moments.

test/common/limit_result.result
```diff
@@ -0,0 +1,6 @@
+Shard 2x2
+RESULT false
```
Totktonada: Don’t sure `false` is expected result. Can you confirm that, please?

graphql/accessor_general.lua
```diff
@@ -631,8 +635,10 @@ end
 --- @tparam table args table of arguments passed within the query except ones
 --- that forms the `filter` parameter
 ---
+--- @tparam table info extra information related to query, passed from core
```
Totktonada: The comment is too general, eps. for read-write paramenter. Moreover, I prefer to describe fields I use (and how) when only part of the table is needed inside a function.

graphql/accessor_general.lua
Totktonada: I would be good to assert on the type of `info` parameter likewise it done for `self` and other parameters.

graphql/accessor_general.lua
```diff
@@ -877,8 +885,14 @@ function accessor_general.new(opts, funcs)
                     'from must be nil or from.connection_name ' ..
                     'must be a string, got ' ..
                     type((from or {}).connection_name))
+                -- use `info.qcontext` to store pre-query variables
+                info.qcontext.stats = info.qcontext.stats or {
+                    objects_cnt = 0,
+                    objects_fetched = 0
+                }
+                info.qcontext.objects_cnt_max = objects_cnt_max
```
Totktonada: Why do you store it per-query, not within `self`?

graphql/accessor_general.lua
```diff
@@ -877,8 +885,14 @@ function accessor_general.new(opts, funcs)
                     'from must be nil or from.connection_name ' ..
                     'must be a string, got ' ..
                     type((from or {}).connection_name))
+                -- use `info.qcontext` to store pre-query variables
```
Totktonada: typo: pre → per

graphql/accessor_general.lua
Totktonada: I think that you should add limit of fetched objects too, otherwise the counter will be useless and just increase confusing when the code will be read.

graphql/core/execute.lua
```diff
@@ -88,7 +88,10 @@ local function buildContext(schema, tree, rootValue, variables, operationName)
     rootValue = rootValue,
     variables = variables,
     operation = nil,
-    fragmentMap = {}
+    fragmentMap = {},
+    -- The field is passed to resolve within info attribute.
```
Totktonada: > The field is passed to resolve within info attribute.

Sorry, but I can’t understand what do you mean by the word 'resolve' here?

graphql/accessor_general.lua
```diff
@@ -837,6 +844,7 @@ function accessor_general.new(opts, funcs)
     local collections = opts.collections
     local service_fields = opts.service_fields
     local indexes = opts.indexes
+    local objects_cnt_max = opts.objects_cnt_max or OBJECT_CNT_MAX
```
Totktonada: You should update ldoc comment above to reflect added / changed / deleted options. In that case you should add the new item to the 'example of … item` list, I guess.

Totktonada (commented):
A bit more comments. General note: assert on parameters (at least where other parameters are verified with asserts) and document API / parameters changes (ldoc).

graphql/accessor_general.lua
```diff
@@ -604,10 +606,12 @@ local function process_tuple(state, tuple, opts)
     -- add the matching object, update count and check limit
     state.objs[#state.objs + 1] = obj
     state.count = state.count + 1
+    stats.objects_cnt = stats.objects_cnt + 1
+    assert(stats.objects_cnt <= objects_cnt_max,
```
Totktonada: You steal my assert!

The meaning of these two asserts is different. Mine one is self-check that the count of objects is equal to the user-provided limit in case it is reached. Your one is the error to report user we cannot proceed with such request. I think you should not steal my assert, but copy it :)

graphql/accessor_general.lua
```diff
@@ -604,10 +606,12 @@ local function process_tuple(state, tuple, opts)
     -- add the matching object, update count and check limit
     state.objs[#state.objs + 1] = obj
     state.count = state.count + 1
+    stats.objects_cnt = stats.objects_cnt + 1
```
Totktonada: This fragment of code is a bit confusing at the first glance, because of the two reasons:

1. state and stats looks very similar here.
2. count and objects_cnt looks like it have the same meaning.

I propose to rename stats to qstats or query_stats to give a developer more anchors when it read the code.

Sorry that I have such many proposals for renaming. Of course, we must discuss all moments that you feel as nitpicking. Really, send me a message if you don’t feel comfortable with some of these proposals.

test/common/limit_result.test.lua
Totktonada: Why 'td'? Abbreviations don’t make your code looks better…

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,75 @@
...
+    print(('RESULT %s'):format(contains(result, "count[4] exceeds limit[3]")))
```
Totktonada: Maybe you want to use test:like() from our tap module? Just guessing…

test/common/lua/td_user_order.lua
```diff
@@ -0,0 +1,139 @@
...
+        }})
+]]
+
+testdata.drop_spaces_script = [[
+    box.space.user_collection:drop()
+    box.space.order_collection:drop()
+]]
+
+
+function testdata.fill_test_data(virtbox)
```
Totktonada: More re unnecessary abbreviations: 'virt' is not better than 'virtual'. The parameter can be named `box_space` if you want.

test/common/lua/td_user_order.lua
```diff
@@ -0,0 +1,139 @@
...
+                index_type = 'tree',
+                unique = false,
+                primary = false,
+            },
+        },
+    }
+}
+
+
+testdata.init_spaces_script = [[
```
Totktonada: I dislike the idea to have a code written as a string. I understand you want to send it using netbox’s eval. Maybe string.dump does the work? If so, can you reuse the test/testdata/common_testdata.lua file instead of copying or you dislike the API I use for such test data files? Maybe we must document the API and validate testdata modules against it, does that make you more happy? :)

Totktonada:
Because I have many comments re things of different importancy I’ll summarize what I want to be resolved (fixed or discussed) before landing to the master branch:

1. Leak of 'executor' abstractions with passing the `info` parameter (`qcontext` passing was proposed).
2. Naming of qcontext’s fields (and consistency with accessor_general opts).
3. Add the second limit (`fetched`) or remove corresponding counter.
4. Ldoc comments updates where an API changed.
5. Reuse the test data if possible, don’t copy.
6. (desirable) Luacheck for all lua code except *.test.lua within the `core = tarantool` tests.

The rest is up to you. I don’t want to make code reviews feels as excruciations.

test/common/suite.ini
```diff
@@ -0,0 +1,5 @@
+[default]
+core = app
+description = tests different setups simultaneously
+lua_libs = lua/multirunner.lua lua/td_user_order.lua
```
Totktonada: Please, consider explicit require (after the fio+path trick) as an alternative to adding a test data to lua_libs.

By the way, the test cannot be run w/o multirunner, just as the lua file?

graphql/accessor_general.lua
Khatskevich: Describe info in more detail in documentation

Totktonada: Discussed and aggreed that we can pass `extra` parameter with `qcontext` only for now. Need to be documented properly.

test/common/limit_result.test.lua
Totktonada: Okay, I do the same in my app tests (but use background = false).

test/common/limit_result.test.lua
Totktonada: require('tap')

test/common/lua/td_user_order.lua
Totktonada: `box_space` is readed as only space, not shard. Virtbox is okay.

test/common/suite.ini
Totktonada: Decided to postpone decision how we should write tests.

test/common/limit_result.test.lua
Khatskevich: background is false by default

.luacheckrc
```diff
@@ -2,7 +2,8 @@ std = {
     read_globals = {'box', 'require', 'debug', 'pcall', 'xpcall', 'tostring',
         'tonumber', 'type', 'assert', 'ipairs', 'math', 'error', 'string',
         'table', 'pairs', 'os', 'select', 'unpack', 'dofile', 'next',
-        'getmetatable', 'setmetatable', 'rawget', 'print'
+        'getmetatable', 'setmetatable', 'rawget', 'print', 'shard_status',
```
Totktonada, line 5: [minor] IMO, it is better to use shard.servers (where `shard` is `require('shard')` value) to avoid global variables.

.luacheckrc
```diff
@@ -2,7 +2,8 @@ std = {
...
+	'loadstring'
```
Totktonada, line 6: [minor] You are use tab here instead of 8 spaces. Please, keep consistency within at least one file and better within the whole project (except Makefiles).

Makefile
```diff
@@ -4,6 +4,7 @@ default:
 .PHONY: lint
 lint:
 	luacheck graphql/*.lua test/local/*.lua test/testdata/*.lua \
+		test/common/*test.lua test/common/lua/*.lua \
```
Totktonada, line 7: [minor] I would add `test/common/*.lua`, `test/common/lua/*.lua` or at least fix `*test.lua` with `*.test.lua` to avoid any possible confusion.

graphql/accessor_general.lua
```diff
@@ -580,6 +585,14 @@ local function process_tuple(state, tuple, opts)
     local filter = opts.filter
     local do_filter = opts.do_filter
     local pivot_filter = opts.pivot_filter
+    local qstats = state.statistics
+    local resulting_object_cnt_max = opts.resulting_object_cnt_max
+    local fetched_object_cnt_max = opts.fetched_object_cnt_max
+    qstats.fetched_object_cnt = qstats.fetched_object_cnt + 1
+    assert(qstats.fetched_object_cnt <= fetched_object_cnt_max,
+            ('fetched object count[%d] exceeds limit[%s] '..
```
Totktonada, line 593: [minor]
1. %s -> %d
2. set `fetched_object_cnt_max` in accessor -> `fetched_object_cnt_max` in accessor [that is the name of the limit, yep?]

Totktonada: 3. Separate string concatenation operation (as well as any other) with whitespaces, this would fit our lua style guidelines better.

graphql/accessor_general.lua
```diff
@@ -604,10 +617,12 @@ local function process_tuple(state, tuple, opts)
     -- add the matching object, update count and check limit
     state.objs[#state.objs + 1] = obj
     state.count = state.count + 1
+    qstats.resulting_object_cnt = qstats.resulting_object_cnt + 1
+    assert(qstats.resulting_object_cnt <= resulting_object_cnt_max,
+            ('returning object count[%d] exceeds limit[%s] '..
```
Totktonada, line 622: [minor] The same comments as above for fetched objects counter.

graphql/accessor_general.lua
```diff
@@ -791,7 +812,10 @@ end
 ---
 --- @tparam table opts `schemas`, `collections`, `service_fields` and `indexes`
 --- to give the data accessor all needed meta-information re data; the format is
---- shown below
+--- shown below; additional attributes `resulting_object_cnt_max` and
+--- `fetched_object_cnt_max` are optional natural numbers whish help to control
```
Totktonada, line 816: [minor]

1. natural -> positive (excluding zero)?
2. whish -> which?

graphql/accessor_general.lua
```diff
@@ -791,7 +812,10 @@ end
...
+--- query behaviour in case it requires more resources than expected _(default
+--- value is 10.000)_
```
Totktonada, line 818: [minor]
3. English uses comma as the separator within a number, not a period.

graphql/accessor_general.lua
```diff
@@ -9,6 +9,8 @@ local avro_schema = require('avro_schema')
 local utils = require('graphql.utils')
 
 local accessor_general = {}
+local RESULTING_OBJECT_CNT_MAX = 10000
+local FETCHED_OBJECT_CNT_MAX = 10000
```
Totktonada, line 13: [minor] It would be glab to add DEF/DEFAULT suffix/prefix to give a reader hint that is is just defaults that can be redefined (via accessor's options). Yep, I understand that the names becomes long.

graphql/tarantool_graphql.lua
```diff
@@ -517,6 +522,11 @@ end
 ---                 --     destination_args_values = <...>,
 ---                 -- }
 ---                 --
+---                 -- extra is a table which contains additional data for the
+---                 -- query; by now it consists of a single qcontext table,
+---                 -- which can be used by accessor to store any qury-related
```
Totktonada, line 527: [minor] qury -> query

test/common/limit_result.test.lua
```diff
@@ -0,0 +1,86 @@
...
+    })
+
+    local gql_wrapper = graphql.new({
+        schemas = data.meta.schemas,
+        collections = data.meta.collections,
+        accessor = accessor,
+    })
+    data.fill_test_data(virtbox)
+    local query = [[
+    query object_result_max($user_id: Int, $description: String) {
```
Totktonada, line 45: [minor] I would add one indent level here (for the string within `[[` and `]]`).

test/common/lua/multirunner.lua
```diff
@@ -0,0 +1,103 @@
...
+        shard.init(config)
+    end)
+    shard.wait_connection()
+end
+
+local function shard_cleanup(test_run, servers)
+    test_run:drop_cluster(servers)
+    -- crutch!!
+    -- test_run.lua do not delete servers
+    -- todo: should be fixed after in tesr-run
```
Totktonada, line 25: [minor] The link to the issue / PR or even temporary branch would be good here.

Totktonada (commented):
LGTM. All comments are minor. I think the PR is ready to land to the master.

Makefile
Khatskevich: `test/common/*.lua` would match `shard[1-4].lua` whish contain warnings (do not remember if warnings minor or not)

Totktonada: Ok, skip it for now.

Khatskevich:
already in master

Issue #39, opened 2018-02-05
JOIN semantic for nested objects filtering


Totktonada:
There are two approach to handle nested objects that are failed to fetch by a 1:1 connection:

* Give an error. This is the current behaviour. Seems to be GraphQL-way.
* Remove all parent objects while parent-child connections has 1:1 type (a first 1:N connected parent or top-level object will give the empty list). This is sort of backtraking and feels like SQL JOIN.

I don’t sure which option should be implemented (or both?).

Totktonada:
Proposal for minimal implementation:

```
query get_orders_by_user_id($user_id: String) {
    order(user: {id: $user_id})
        id
        description
        user {
            id
            first_name
            last_name
        }
    }
}
```

- [ ] tarantool_graphql: add an argument for each 1:1 connection with its arguments as the type: type of the `user` argument above is the same as `object_args` for the order collection (which is destination collection for the user→order connection);
- [ ] accessor_general: detect 'user' field in the filter, fetch and unflat the user for each order (w/o any filters, by a connection, just call select_internal recursively), add the user for each order, pass the constructed object to is_subtable as before (it should support nested matching).

Khatskevich:
An idea for optimization: if a condition is set on connection part, then it can be checked in the object itself.

Khatskevich:
## Proposal: Reuse executor for objects fetching
The idea of the proposal: it is not good to implement executor while we have one.
### Implementation notes
In resolve function we can decide if a filter has constraints on inner objects, then it:
1. fetches its own object
2. creates callback, which is called by executor on the whole object when it's fetching is completed
#### Pros
1. do not implement the second executor
2. do not fetch objects two times
3. butching (ant other) optimizations should be implemented only once (in executor)
#### Cons
1. If we do not change executor much, then it will fetch only fields which are mentioned in inner query
```
user (order_connection:{price:3}) {
     user_id
     order_connection {
           description
     }
}
```
will not work, because we have condition on price, but we do not retrieve price
2. The same thing with conditions on nested objects which are not fetched at all
```
user (order_connection:{price:3}) {
     user_id
}
```

Totktonada:
I’m going to implement the first approach (subrequests on the accessor_general level, no even caching) while we have no strong opinion how the future optimized executor will be implemented.

Totktonada:
Note: The PoC implementation (see referencing commit above) does not follow the initial proposal in details. TBD: describe the approach of PoC.

[...]

Issue #13, opened 2018-01-26
Add built-in filter function for every Avro class


kostja:
like()
regexp()
>
<
or
and

Totktonada:
Plan:

- parsing of a filter (string or input object?)
- pcre
- eq, ne, lt, gt, le, ge
- and, or, not
- speedup using indexes

Totktonada:
More from #43:

* IS NULL / IS NOT NULL

Totktonada:
More from #39:

* [maybe] Support for nested filtering by connection_name.field_name.

Totktonada:
Sink down to prio2 since PCRE support was extracted to #73.

Totktonada:
## Syntax

C-style expressions. Field names (dot-separated for inner fields) of a current
object can be used. Variables can be used.

### Grammar

TBD.

## Overview

Create AST using lulpeg in the compile phase, save it in a query AST.

Set an usage flag for variables in use (to avoid an error re non-used
variable).

Pass the AST to accessor_general within `extra` arguments. Implement a tuple
filtering according by the AST in `process_tuple`.

If no index is found based on filters, offset or from a connection (see
`get_index_name`), then it is worth to use the AST to perform index search.
Basically we need to collect all always-checked predicates (maybe it worth to
construct CNF?): ones that are not under OR operator. Then prioritize it: 
equality first, then less-great. Then attempt to match each predicate against
existing indexes (by a name of the field it operates on). When a match found,
then rewrite AST to avoid redundant operations in `process_tuple` (remove the 
operation that will be performed by an index).

The approach above is very-very basic and cannot handle many cases. But I
think it is on optimal point between efficiency and simplicity for now.

## Follow ups 

### Type checking

Each operation argument should have set of allowed types. Say, 'number' for a
+, -, <, >, 'boolean' for ||, &&, !, 'number' and 'string' for == and != as
well as a type of result.

A root expression must have the boolean type, because it will be used as the 
predicate todecided whether an object should be filtered in. 

It seems that we should take care of nullability of a type: say, `regexp` will
not work with null as an argument, but `is_null` will do.

### Functions

E. g. regexp(<pattern: string expr>, <string expr>).

### IS NULL / IS NOT NULL operations

is_null(<expr>)
is_not_null(<expr>)

PR #48, opened 2018-02-21
[WIP] Sb/avro arr

### LINKED ISSUE #4: Support for avro array / map
- [x] Generate GraphQL schema for arrays and maps.
- [x] Remove array / map fields from arguments list.
- [x] Write test.

SudoBobo:
closes #4


graphql/tarantool_graphql.lua
```diff
@@ -49,16 +57,7 @@ local function avro_type(avro_schema)
...
-local function nullable(gql_class)
-    assert(type(gql_class) == 'table', 'gql_class must be a table, got ' ..
-        type(gql_class))
-
-    if gql_class.__type ~= 'NonNull' then return gql_class end
-
-    assert(gql_class.ofType ~= nil, 'gql_class.ofType must not be nil')
-    return gql_class.ofType
-end
+local nullable = utils.nullable
```
Totktonada: utils.lua is for auxiliary functions that do not contain domain-specific abstractions or logic. We work with graphql-lua types in core/*.lua and tarantool_graphql.lua. It is better to keep these objects within these files (or create new modules working within such terms), but don't mix abstraction layers. In other words, I propose to give back this function to this module. You can create graphql_utils.lua alternatively, but I don't think we have enough helpers to charge it.

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +95,25 @@ local function convert_scalar_type(avro_schema, opts)
     elseif avro_t == 'string*' then
         return types.string
     end
+
+    if opts.is_items_type then
```
Totktonada: // This is not more actual because I proposed to remove this option, but I kept this comment here to give you more on our conventions within the project.

Please, keep consistency with asserts and default values for options within the 'opts' parameter. Consider 'raise' as the example. More specific: add the local variable with a default value (`or false`), add assert for a type of this option. This is more about readability, then about runtime checks, because such function preamble give a developer information re function contract.

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +95,25 @@ local function convert_scalar_type(avro_schema, opts)
...
+        error('avro array items must have know scalar type, not: ' ..
+                json.encode(avro_schema))
```
Totktonada: We use 4 spaces as indent.

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +95,25 @@ local function convert_scalar_type(avro_schema, opts)
     elseif avro_t == 'string*' then
         return types.string
     end
+
+    if opts.is_items_type then
+        error('avro array items must have know scalar type, not: ' ..
```
Totktonada: A user must have info to distinguish not-implemented-yet errors from its mistakes in configuration options (like avro-schemas).  Proposed wording: "only scalars are supported as array items for now".

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +95,25 @@ local function convert_scalar_type(avro_schema, opts)
...
+                json.encode(avro_schema))
+    end
+
     if raise then
         error('unrecognized avro-schema scalar type: ' ..
-            json.encode(avro_schema))
+        json.encode(avro_schema))
     end
+
```
Totktonada: We don't use python-style double empty lines (moreover, here they are within a function, so don't fit PEP8 too). Consider our lua style guidelines: https://tarantool.org/en/doc/1.7/dev_guide/lua_style_guide.html

graphql/tarantool_graphql.lua
Totktonada: Don't sure we should pass this option to this function, because it is out of the scope of converting scalar types (this is what the function does). By the way, this option breaks the `raise` option behavior (now it cannot be described just as '`raise = true` let the function raise an error on a wrong scalar avro type, `raise = false` let the function return `nil` in the case'). I have strong opinion that you should process this situation in a callee code.

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +95,25 @@ local function convert_scalar_type(avro_schema, opts)
...
+    if opts.is_items_type then
+        error('avro array items must have know scalar type, not: ' ..
+                json.encode(avro_schema))
+    end
+
     if raise then
         error('unrecognized avro-schema scalar type: ' ..
-            json.encode(avro_schema))
+        json.encode(avro_schema))
```
Totktonada: Hmm. Cannot get you. Why do you remove the indent? Now it feels as broken.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
+        if not is_for_args or (nullable(res[field.name].kind) ~= 'List'
+                            and nullable(res[field.name].kind) ~= 'Map') then
+            object_args[field.name] = nullable(res[field.name].kind)
+        end
     end
     return res, object_args
 end
 
---- The function recursively converts passed avro-schema to a graphql type.
+--- The function recursively converts passed avro-schema to a graphql type (kind).
```
Totktonada: Personally I keep my lines within 79 symbols. The limit can be discussed.

graphql/tarantool_graphql.lua
```diff
@@ -115,17 +124,21 @@ local function gql_argument_type(state, avro_schema)
...
             :format(type(avro_schema.name), json.encode(avro_schema)))
+
         assert(type(avro_schema.fields) == 'table',
             ('avro_schema.fields must be a table, got %s (avro_schema %s)')
             :format(type(avro_schema.fields), json.encode(avro_schema)))
 
         local fields = {}
-        for _, field in ipairs(fields) do
+        for _, field in ipairs(avro_schema.fields) do
+
```
Totktonada: Feel free to use single empty line as the separator for code blocks, but I propose to don't separate a loop header from its body (or a first code block of the body). The same for empty lines before / after `else` / `elseif` / `end`.

If you have some certain view how we should change our code style, let discuss it first, then (maybe) make changes.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
     end
 end
 
+
+--- Recursively convert each field of an avro-schema to a graphql type and
+--- corresponding argument for an upper graphql type.
+---
+--- @tparam table state
```
Totktonada: Add XXX comment if you make some kind of placeholder or partial functionality / documentation, but don't finish it in some manner. Here I mean comments for parameters of the function. If you describe it in some other place, you can use `@{...}` ldoc syntax to point reader to the right place to get info re a parameter.

graphql/tarantool_graphql.lua
```diff
@@ -115,17 +124,21 @@ local function gql_argument_type(state, avro_schema)
...
             ('avro_schema.name must be a string, got %s (avro_schema %s)')
             :format(type(avro_schema.name), json.encode(avro_schema)))
+
         assert(type(avro_schema.fields) == 'table',
             ('avro_schema.fields must be a table, got %s (avro_schema %s)')
             :format(type(avro_schema.fields), json.encode(avro_schema)))
 
         local fields = {}
-        for _, field in ipairs(fields) do
+        for _, field in ipairs(avro_schema.fields) do
```
Totktonada, line 152: My bad. Thanks for pointing it out! Filed #53 for that.

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +95,25 @@ local function convert_scalar_type(avro_schema, opts)
...
     end
+
+
     return nil
 end
 
 --- Non-recursive version of the @{gql_type} function that returns
 --- InputObject instead of Object.
+--- An error will be raised in case of fields that are not scalar types
+--- as there are no sense in non scalar arguments.
```
Totktonada: 1. You described the case when a record was passed as a function parameter, state that explicitly.
2. The record case *is* about a non-scalar argument, we are support that for offset when a primary key is compound. We'll support more non-scalar arguments soon (likely for nested filtering from #39, maybe for complex filters from #13).
3. Try being as specific as you can when writing a documentation. 'There are no sense' comment has no sense itself in the most cases: a) when it is obvious; b) when it is not obvious, but there are no mention of the certain reason.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
+
+--- Recursively convert each field of an avro-schema to a graphql type and
+--- corresponding argument for an upper graphql type.
+---
+--- @tparam table state
+--- @tparam table fields
 local function convert_record_fields_to_args(state, fields)
     local args = {}
     for _, field in ipairs(fields) do
+
```
Totktonada: Extra empty line after the `for` header, check the related comment above.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
     end
 end
 
+
+--- Recursively convert each field of an avro-schema to a graphql type and
+--- corresponding argument for an upper graphql type.
```
Totktonada: 1. Non-recursively.
2. The wording 'graphql type and argument' sounds confusing for me (expected two arguments). Proposed wording is below.

> Convert list of fields in the avro-schema format to list of GraphQL types with intention to use it as GraphQL arguments later.
>
> It uses the @{gql_argument_type} function to convert each field, then skips fields of array and map types and gives the resulting list of converted fields.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
         assert(type(field.name) == 'string',
             ('field.name must be a string, got %s (schema %s)')
             :format(type(field.name), json.encode(field)))
+
         local gql_class = gql_argument_type(state, field.type)
-        args[field.name] = nullable(gql_class)
+
+        -- arrays (gql lists) and maps can't be arguments
+        -- so these kinds are to be skipped
+
```
Totktonada: Don't separate a code block comment from its code block with empty line.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
         assert(type(field.name) == 'string',
             ('field.name must be a string, got %s (schema %s)')
             :format(type(field.name), json.encode(field)))
+
         local gql_class = gql_argument_type(state, field.type)
-        args[field.name] = nullable(gql_class)
+
+        -- arrays (gql lists) and maps can't be arguments
+        -- so these kinds are to be skipped
```
Totktonada: I don't think we should use the term 'kind' in sense 'a graphql type in the graphql-lua format'. We use 'graphql type' term for that already.

General note: I suggest to try hard to use one term for an one abstraction with a project and use different terms for different abstractions.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
+
+        if (nullable(gql_class) ~= 'List'
```
Totktonada: Don't add extra parenthesis around an `if` condition (it is not required by the Lua syntax).

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
+                            and nullable(gql_class) ~= 'Map') then
```
Totktonada: Indent here is strange. Use double indent (8 spaces) when you need to carry the line with an `if` condition.

graphql/tarantool_graphql.lua
Totktonada: As I write above I think we need to omit usage of the term 'kind' for graphql types.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
+            args[field.name] = nullable(gql_class)
+        end
     end
     return args
 end
 
---- Convert each field of an avro-schema to a graphql type and corresponding
---- argument for an upper graphql type.
+--- Recursively convert each field of an avro-schema to a graphql type and
```
Totktonada: I think that such implementation details is not for the header string of the function. You can write details below, like so:

```
--- What the function does.
---
--- Any details here.
---
--- ...params and return values...
local function...
```

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
         res[field.name] = {
             name = field.name,
             kind = gql_type(state, field.type),
         }
-        object_args[field.name] = nullable(res[field.name].kind)
+
+
+        -- arrays (gql lists) and maps can't be arguments
+        if not is_for_args or (nullable(res[field.name].kind) ~= 'List'
+                            and nullable(res[field.name].kind) ~= 'Map') then
```
Totktonada: Use double indent for such cases.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
             ('field.name must be a string, got %s (schema %s)')
             :format(type(field.name), json.encode(field)))
+
         res[field.name] = {
             name = field.name,
             kind = gql_type(state, field.type),
         }
-        object_args[field.name] = nullable(res[field.name].kind)
+
+
```
Totktonada: Use single empty line to split code blocks.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
         if res == nil then
             error('unrecognized avro-schema type: ' .. json.encode(avro_schema))
         end
         return res
+
     end
+
 end
 
+
```
Totktonada: Don't use python-style double empty lines.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
+            'top-level schema must have record avro type, not' .. schema.type)
+        state.types[collection_name] = gql_type(state, schema, collection, collection_name)
+
+        -- prepare arguments (their kinds)
         local _, object_args = convert_record_fields(state,
-            schema.fields)
+            schema.fields, {is_for_args=true})
         local list_args = convert_record_fields_to_args(
-            state, accessor:list_args(name))
+            state, accessor:list_args(collection_name), {is_for_args=true})
```
Totktonada: 1. You pass a parameter that you don't define in the function profile.
2. We must not filter accessor-provided arguments. Feel free to add asserts that all arguments we got here are correct.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
---- Convert each field of an avro-schema to a graphql type and corresponding
---- argument for an upper graphql type.
+--- Recursively convert each field of an avro-schema to a graphql type and
+--- corresponding argument for an upper graphql type.
 ---
 --- @tparam table state for read state.accessor and previously filled
 --- state.types
 --- @tparam table fields fields part from an avro-schema
-local function convert_record_fields(state, fields)
+--- @tparam table opts include is_for_args flag to specify
```
Totktonada: The parameter is redundant. You always set it to `true` when use the `object_args` result. Just do true-case action always.

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +162,71 @@ local function gql_argument_type(state, avro_schema)
...
+
         res[field.name] = {
             name = field.name,
             kind = gql_type(state, field.type),
         }
-        object_args[field.name] = nullable(res[field.name].kind)
+
+
+        -- arrays (gql lists) and maps can't be arguments
+        if not is_for_args or (nullable(res[field.name].kind) ~= 'List'
```
Totktonada: IMO, it is better to use `avro_type(field.type)` here. Maybe I think so because I don't know exact structure of graphql-lua types implementation.

graphql/tarantool_graphql.lua
```diff
@@ -261,20 +305,24 @@ gql_type = function(state, avro_schema, collection, collection_name)
 
             local c_list_args = state.list_arguments[c.destination_collection]
 
+            -- change fields that are represented by connections
```
Totktonada: That is not about changing, here we just add fields by connections.

graphql/tarantool_graphql.lua
```diff
@@ -261,20 +305,24 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
             fields[c.name] = {
                 name = c.name,
                 kind = destination_type,
                 arguments = c_args,
                 resolve = function(parent, args_instance, info)
                     local destination_args_names = {}
                     local destination_args_values = {}
+
                     for _, part in ipairs(c.parts) do
+
```
Totktonada: Extra empty line after the `for` header.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
             }
         end
 
+        -- create gql schema
```
Totktonada: schema -> type

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
         local res = types.object({
             name = collection ~= nil and collection.name or avro_schema.name,
             description = 'generated from avro-schema for ' ..
                 avro_schema.name,
             fields = fields,
         })
         return avro_t == 'enum' and types.nonNull(res) or res
+
```
Totktonada: Extra empty line before `elseif`.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
     elseif avro_t == 'enum' then
         error('enums not implemented yet') -- XXX
+
```
Totktonada: Extra empty line before `elseif`.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+    elseif avro_t == 'array' or avro_t == 'array*' then
+
```
Totktonada: Extra empty line after`elseif`.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+        assert(avro_schema.items ~= nil,
+            'items field must not be nil in array avro schema')
+        assert(type(avro_schema.items) == 'string',
+            'avro_schema.items must be a string, got ' .. type(avro_schema.item))
```
Totktonada: The line is longer than 79 symbols.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+        local gql_array = types.list(gql_items_type)
+
+        if  avro_t == 'array*' then
+            return gql_array
+        end
+
+        if avro_t == 'array' then
+            return types.nonNull(gql_array)
+        end
+
```
Totktonada: Extra empty line before `else`.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+            'items field must not be nil in array avro schema')
+        assert(type(avro_schema.items) == 'string',
+            'avro_schema.items must be a string, got ' .. type(avro_schema.item))
+
+        local gql_items_type = convert_scalar_type(avro_schema.items,
+            {is_items_type=true, raise=true})
+
+        local gql_array = types.list(gql_items_type)
+
+        if  avro_t == 'array*' then
```
Totktonada: 1. Double space after `if`.
2. IMO that two `if`s would look better with Lua's 'ternary operator': `return avro_t == 'array' and types.nonNull(gql_array) or gql_array`.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+
+    elseif avro_t == 'array' or avro_t == 'array*' then
+
+        assert(avro_schema.items ~= nil,
+            'items field must not be nil in array avro schema')
+        assert(type(avro_schema.items) == 'string',
+            'avro_schema.items must be a string, got ' .. type(avro_schema.item))
+
+        local gql_items_type = convert_scalar_type(avro_schema.items,
+            {is_items_type=true, raise=true})
```
Totktonada: You duplicate logic of the raise flag with the is_items_type flag, yep? I don't get why. If you want to make custom error message then use `raise = false` and assert for `nil` result, I use that approach with this function (grep it).

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
         error('enums not implemented yet') -- XXX
+
+    elseif avro_t == 'array' or avro_t == 'array*' then
+
+        assert(avro_schema.items ~= nil,
+            'items field must not be nil in array avro schema')
+        assert(type(avro_schema.items) == 'string',
+            'avro_schema.items must be a string, got ' .. type(avro_schema.item))
+
+        local gql_items_type = convert_scalar_type(avro_schema.items,
```
Totktonada: I think you need to leave XXX comment here that it is not clear what to do with complex types inside arrays (just pass to results or allow to use filters and so on), so we allow only scalar arrays for now.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+            return types.nonNull(gql_array)
+        end
+
     else
         local res = convert_scalar_type(avro_schema, {raise = false})
         if res == nil then
             error('unrecognized avro-schema type: ' .. json.encode(avro_schema))
         end
         return res
+
```
Totktonada: Extra empty line before `end`.

graphql/tarantool_graphql.lua
```diff
@@ -318,24 +366,50 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
     end
+
```
Totktonada: Extra empty line before `end`.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
             tostring(collection.schema_name)))
         assert(schema.name == nil or schema.name == collection.schema_name,
             ('top-level schema name does not match the name in ' ..
             'the schema itself: "%s" vs "%s"'):format(collection.schema_name,
             schema.name))
-        state.types[name] = gql_type(state, schema, collection, name)
 
+        -- recursively converts all avro types into gql types in the given schema
+        assert(schema.type == 'record',
+            'top-level schema must have record avro type, not' .. schema.type)
```
Totktonada: Space is missing after 'not'.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
+        state.types[collection_name] = gql_type(state, schema, collection, collection_name)
```
Totktonada: >79 symbols.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
+
+        -- prepare arguments (their kinds)
         local _, object_args = convert_record_fields(state,
-            schema.fields)
+            schema.fields, {is_for_args=true})
```
Totktonada: Proposed to remove the parameter (see above).

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
             'the schema itself: "%s" vs "%s"'):format(collection.schema_name,
             schema.name))
-        state.types[name] = gql_type(state, schema, collection, name)
 
+        -- recursively converts all avro types into gql types in the given schema
+        assert(schema.type == 'record',
+            'top-level schema must have record avro type, not' .. schema.type)
+        state.types[collection_name] = gql_type(state, schema, collection, collection_name)
+
+        -- prepare arguments (their kinds)
```
Totktonada: Proposed to don't use the term 'kind' for graphql types (see above).

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
         local list_args = convert_record_fields_to_args(
-            state, accessor:list_args(name))
+            state, accessor:list_args(collection_name), {is_for_args=true})
         local args = utils.merge_tables(object_args, list_args)
-        state.object_arguments[name] = object_args
-        state.list_arguments[name] = list_args
-        state.all_arguments[name] = args
+
+        -- list and map (avro array and map) can't be arguments
+
```
Totktonada: Don't separate a code block comment from it code blocks by an empty line.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +430,41 @@ local function parse_cfg(cfg)
...
+            schema.fields, {is_for_args=true})
         local list_args = convert_record_fields_to_args(
-            state, accessor:list_args(name))
+            state, accessor:list_args(collection_name), {is_for_args=true})
         local args = utils.merge_tables(object_args, list_args)
-        state.object_arguments[name] = object_args
-        state.list_arguments[name] = list_args
-        state.all_arguments[name] = args
+
+        -- list and map (avro array and map) can't be arguments
```
Totktonada: I don't get, how this comment related to the code above? Maybe we need to move it to the right place.

graphql/tarantool_graphql.lua
```diff
@@ -414,6 +497,8 @@ local function parse_cfg(cfg)
     return state
 end
 
+--- The function checks if given query has an appropriate type 'query'
```
Totktonada: It seems that the function does the other action: it checks that one and only one GraphQL operation (query/mutation/subscription) is defined in the AST. That just the minimum check that we got an query AST and not mistakely some other value.

graphql/tarantool_graphql.lua
```diff
@@ -414,6 +497,8 @@ local function parse_cfg(cfg)
...
+--- (mutations are not supported yet).
```
Totktonada: That note is unrelated to the function. It will doing what it doing despite types of GraphQL operations.

graphql/tarantool_graphql.lua
```diff
@@ -442,6 +529,12 @@ local function gql_execute(qstate, variables)
         operation_name)
 end
 
+--- The function parses a raw query string, validate the resulting query
```
Totktonada: 1. The term 'raw' is a bit confusing, I think. Just 'a query string'.
2. 'Validate the resulting query'. Too common. Make it specific by adding 'against the GraphQL scheme.
3. 'Make it ready for execution'. Too common. You can say 'provides an object with the function to execute the query with specific variables values).

graphql/tarantool_graphql.lua
```diff
@@ -442,6 +529,12 @@ local function gql_execute(qstate, variables)
...
+--- and make it ready for execution.
+---
+--- @tparam table state current state of graphql-lib, including
```
Totktonada: We don't used the term 'graphql-lib' before.

graphql/tarantool_graphql.lua
```diff
@@ -442,6 +529,12 @@ local function gql_execute(qstate, variables)
...
+--- schemas, collections and accessor
+--- @tparam string query raw query string
```
Totktonada: Proposed to omit the word 'raw'.

Totktonada (commented):
Style comments, opinions re implementation approaches.

graphql/tarantool_graphql.lua
SudoBobo: 2. So the point here is that currently we do not support triple nesting level for arguments, but do support double nesting level?

graphql/tarantool_graphql.lua
```diff
@@ -79,7 +87,7 @@ local function convert_scalar_type(avro_schema, opts)
     assert(type(opts) == 'table', 'opts must be nil or table, got ' ..
         type(opts))
     local raise = opts.raise or false
-    assert(type(opts.raise) == 'boolean', 'opts.raise must be boolean, got ' ..
+    assert(type(raise) == 'boolean', 'opts.raise must be boolean, got ' ..
         type(opts.raise))
```
Totktonada: Replace `opts.raise` with `raise` here too for consistency.

graphql/tarantool_graphql.lua
Totktonada: Ping?

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +104,21 @@ local function convert_scalar_type(avro_schema, opts)
...
+
     return nil
 end
 
 --- Non-recursive version of the @{gql_type} function that returns
 --- InputObject instead of Object.
+--- An error will be raised if avro_schema type is 'record'
+--- and its' fields are not scalar type because currently
+--- triple nesting level (record with record as a field - ok,
+--- record with record wich has inside another level - not ok)
```
Totktonada: wich → which

graphql/tarantool_graphql.lua
```diff
@@ -149,44 +166,69 @@ local function gql_argument_type(state, avro_schema)
     end
 end
 
+
```
Totktonada: Extra empty line.

graphql/tarantool_graphql.lua
Totktonada: Ping?

graphql/tarantool_graphql.lua
Totktonada: Ping?

graphql/tarantool_graphql.lua
```diff
@@ -197,6 +239,10 @@ end
 ---    automatically generate corresponding decucible fields.
 --- 2. The collection name will be used as the resulting graphql type name
 ---    instead of the avro-schema name.
+---
+--- XXX As it is not clear now what to do with complex types inside arrays
+---(just pass to results or allow to use filters), only scalar arrays
```
Totktonada: Forgotten to add space after `---`.

graphql/tarantool_graphql.lua
```diff
@@ -327,6 +381,19 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+        assert(avro_schema.items ~= nil,
+            'items field must not be nil in array avro schema')
+        assert(type(avro_schema.items) == 'string',
+            'avro_schema.items must be a string, got '
+                .. type(avro_schema.item))
+
+        local gql_items_type = convert_scalar_type(avro_schema.items)
+
+        assert(gql_items_type, "only scalars are supported as array items for now "
+            .. avro_type(avro_schema.items) .. " is not a scalar")
```
Totktonada: It is good to being consistent in using single / double quotes and in `..` at the end / at the start of a line.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +423,41 @@ local function parse_cfg(cfg)
...
             tostring(collection.schema_name)))
         assert(schema.name == nil or schema.name == collection.schema_name,
             ('top-level schema name does not match the name in ' ..
             'the schema itself: "%s" vs "%s"'):format(collection.schema_name,
             schema.name))
-        state.types[name] = gql_type(state, schema, collection, name)
 
+        -- recursively converts all avro types into gql types in the given schema
+        assert(schema.type == 'record',
+            'top-level schema must have record avro type, not'
```
Totktonada: Space after 'not'. I write 'got' in such cases, though.

graphql/tarantool_graphql.lua
```diff
@@ -356,32 +423,41 @@ local function parse_cfg(cfg)
...
+                .. schema.type)
```
Totktonada: I suggest to use `tostring(schema.type)` to correctly show assert in case of lack of the `type` field.

graphql/tarantool_graphql.lua
```diff
@@ -414,6 +490,9 @@ local function parse_cfg(cfg)
     return state
 end
 
+--- The function checks that one and only one GraphQL operation
+--- (query/mutation/subscription) is defined in the AST and it's type
```
Totktonada, line 524: > and it's type is 'query' as mutations and subscriptions are not supported yet

Where do you see this check?

graphql/tarantool_graphql.lua
```diff
@@ -442,6 +523,12 @@ local function gql_execute(qstate, variables)
         operation_name)
 end
 
+--- The function parses a query string, validate the resulting query
+--- against the GraphQL schema and provides an object with the function to
+--- execute the query with specific variables values
```
Totktonada: AFAIR, ldoc expects to having period at end of short description. By the way, I personally prefer to split description (or description paragraphs) and each of params with empty lines (only with `---`).

test/local/avro_array.test.lua
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env tarantool
+
+box.cfg{background = false}
+local fio = require('fio')
+
+-- require in-repo version of graphql/ sources despite current working directory
+package.path = fio.abspath(debug.getinfo(1).source:match("@?(.*/)")
+   :gsub('/./', '/'):gsub('/+$', '')) .. '/../../?.lua' .. ';' ..
+package.path
```
Totktonada: The indent is strange.

test/local/simple_avro_array.test.lua
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env tarantool
+
+local fio = require('fio')
+
+ --require in-repo version of graphql/ sources despite current working directory
+package.path = fio.abspath(debug.getinfo(1).source:match("@?(.*/)")
+                                :gsub('/./', '/'):gsub('/+$', '')) .. '/../../?.lua' .. ';' ..
```
Totktonada: The indent is even stranger then in the test before :)

test/local/simple_avro_array.test.lua
```diff
@@ -0,0 +1,98 @@
...
+                { name = 'limit', type = 'int' },
+                { name = 'offset', type = 'long' },
+            -- {name = 'filter', type = ...},
+            }
+        end,
+    }
+})
+
+local gql_wrapper_simple_accessor = graphql.new({
+-- class_name:class mapping
```
Totktonada: 4 space indent is needed here.

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,113 @@
...
+
+local array_testdata = {}
+
+local function print_and_return(...)
+    print(...)
+    return table.concat({ ... }, ' ') .. '\n'
+end
+
+function array_testdata.get_test_metadata()
+
```
Totktonada: Redundant empty line after the function header.

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,113 @@
...
+    local schemas = json.decode([[{
+    "user": {
```
Totktonada: Need one more indent level.

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,113 @@
...
+        "name": "user",
+        "type": "record",
+        "fields": [
+            { "name": "user_id", "type": "string" },
+            { "name": "favorite_food", "type": {"type": "array", "items": "string"} }
+        ]
+        }
```
Totktonada: Here indent is broken too.

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,113 @@
...
+end
+
+function array_testdata.init_spaces()
+    -- user_collection fields
+    local U_USER_ID_FN = 2
+
+    box.once('test_space_init_spaces', function()
+        box.schema.create_space('user_collection')
+        box.space.user_collection:create_index('user_id_index',
+            { type = 'tree', unique = true, parts = {
```
Totktonada: Minor comment. I’m a bit against such style of curly circles / indent. If the line carry, then do not start it on the same line as open curly bracket.

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,113 @@
...
+    --@todo add empty array
+end
+
+function array_testdata.drop_spaces()
+    box.space._schema:delete('oncetest_space_init_spaces')
+    box.space.user_collection:drop()
+end
+
+function array_testdata.run_queries(gql_wrapper)
+
```
Totktonada: Redundant empty line after the function header.

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,113 @@
...
+    local query_1 = [[
+        query user_favorites($user_id: String) {
+            user_collection(user_id: $user_id) {
+                user_id
+                favorite_food
+            }
+        }
+    ]]
+
+    --assert(false, 'err')
```
Totktonada: Forgotten debug, I guess.

Totktonada (commented):
Some style comments.

Totktonada:
> graphql/utils.lua 100644 → 100755

The module is not desined to be run as executable. I think it is better to don’t have executable bit set on such modules. @SudoBobo, please, revert (`chmod a-x graphql/utils.lua`) the change.

graphql/tarantool_graphql.lua
SudoBobo:
```
assert(ast.definitions[1].operation == 'query',
func_name .. ': expected a query operation')
```
and it actually falls if we try to make something like:
```
mutation userFavs($user_id: String) {
        user_collection(user_id: $user_id) {
            user_id
            favorite_food
        }
    }
```

Totktonada: Ouch, sorry.

test/local/simple_avro_array.test.lua
SudoBobo: Sorry, that was produced by my IDE. Fixed that

test/testdata/array_and_map_testdata.lua
```diff
@@ -0,0 +1,142 @@
...
+    local query = [[
+        query user_holidays($user_id: String) {
+            user_collection(user_id: $user_id) {
+                user_id
+                favorite_food
+                favorite_holidays
+                user_balances {
+                    value
+                }
+                customer_balances
```
Totktonada, line 122: Should it work when you don’t choose any fields inside?

SudoBobo: Yes, because it is Map and our Map is a scalar-based type. In GraphQL scalar-based types don't have any inner fields

Totktonada: Okay, I got it. Thanks! I think we should mention record vs map differences in README.md or somewhere like.

README.md
```diff
@@ -47,6 +47,42 @@ GraphQL queries on data from the local Tarantool's storage called spaces.
...
 
+### Notes on types
+
+User should distinguish between Object and Map types. Both of them consists of
+keys and values but there are some important differences. 
+
+While Object is a GraphQL
+built-in type, Map is a scalar-based type. In case of Object-based type 
+all key-value pairs are set during type definition and values may have different 
+arbitrary types.
```
Totktonada: different arbitrary types → different types as defined in the schema; what do you think?

README.md
```diff
@@ -47,6 +47,42 @@ GraphQL queries on data from the local Tarantool's storage called spaces.
...
+
+In contrast, all values of Map-based type must have the same
+type and specific key-value pairs are not set during type definition.
```
Totktonada: specific key-value pairs are not set during type definition → set of valid keys is not defined in the schema, any key-value pair is valid despite name of the key while value has schema-determined type; what do you think?

README.md
```diff
@@ -47,6 +47,42 @@ GraphQL queries on data from the local Tarantool's storage called spaces.
...
+
+Map-based types should be queried as a scalar type, not as an object type 
+(because map's keys are not part of the schema)
```
Totktonada: Period at the end.

README.md
```diff
@@ -47,6 +47,42 @@ GraphQL queries on data from the local Tarantool's storage called spaces.
...
+
+While Object is a GraphQL
+built-in type, Map is a scalar-based type. In case of Object-based type 
+all key-value pairs are set during type definition and values may have different 
+arbitrary types.
+
+In contrast, all values of Map-based type must have the same
+type and specific key-value pairs are not set during type definition.
+
+Map-based types should be queried as a scalar type, not as an object type
```
Totktonada: * as a scalar type → as a scalar type (strictly w/o set of nested fields);
* as an object type → as an object type (strictly with set of nested fields).

Totktonada (commented):
README comments.

graphql/core/execute.lua
```diff
@@ -79,6 +79,13 @@ local function mergeSelectionSets(fields)
 end
 
 local function defaultResolver(object, arguments, info)
+  --print('print from default resolver in execute 82 ')
+  --print(object)
+  --require('pl.pretty').dump(object)
+  --require('pl.pretty').dump(object)
+  --print('print from default resolver in 86')
+  --require('pl.pretty').dump(info.fieldASTs[1].name.value)
+  --require('pl.pretty').dump(object[info.fieldASTs[1].name.value])
```
Totktonada: Forgotten debug.

graphql/core/execute.lua
```diff
@@ -148,6 +155,25 @@ end
 
 local evaluateSelections
 
+--- check if given object is flat and have only scalars
```
Totktonada: If you want to add ldoc comment, then use it's convention: start short function description with a capital letter and end it with a period. Otherwise, use a regular comment (`--`).

graphql/tarantool_graphql.lua
```diff
@@ -72,15 +80,29 @@ local types_long = types.scalar({
     end
 })
 
+local types_map = types.scalar({
+    name = 'Map',
+    description = 'Map is a dictionary with string keys and values of ' ..
+        'arbitrary but same among all values type',
+    serialize = function(value) return value end,
+    parseValue = function(value)return value end,
+    -- node == ast
```
Totktonada, line 89: I think you should comment what is purpose of this line of code or remove the line completely.

README.md
SudoBobo: Your variant is more specific.

graphql/tarantool_graphql.lua
```diff
@@ -96,15 +118,21 @@ local function convert_scalar_type(avro_schema, opts)
...
     end
+
     return nil
 end
 
 --- Non-recursive version of the @{gql_type} function that returns
 --- InputObject instead of Object.
+--- An error will be raised if avro_schema type is 'record'
+--- and its' fields are not scalar type because currently
+--- triple nesting level (record with record as a field - ok,
```
Totktonada, line 134: Currently triple nesting level... what? Is not supported?

graphql/tarantool_graphql.lua
```diff
@@ -197,6 +251,10 @@ end
 ---    automatically generate corresponding decucible fields.
 --- 2. The collection name will be used as the resulting graphql type name
 ---    instead of the avro-schema name.
+---
+--- XXX As it is not clear now what to do with complex types inside arrays
+--- (just pass to results or allow to use filters), only scalar arrays
```
Totktonada, line 256: You can mention here that map is included into scalar types for now.

README.md
SudoBobo: set of valid keys is not defined in the schema, any key-value pair is valid despite name of the key while value has schema-determined type (which is the same among all values in the map) ?

graphql/tarantool_graphql.lua
```diff
@@ -329,10 +395,34 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+    elseif avro_t == 'map' or avro_t == 'map*' then
+        assert(avro_schema.values ~= nil,
+            'values must not be nil in map avro schema')
+        assert(type(avro_schema.values) == 'table'
+            or type(avro_schema.values) == 'string',
+            ('avro_schema.values must be a table or a string,' ..
+            'got %s (avro_schema %s)'):format(type(avro_schema.values),
+            json.encode(avro_schema)))
+
+        gql_type(state, avro_schema.values)
```
Totktonada, line 418: Do you want to comment that it is for asserts? Just to validate the values format against avro-schema format (

README.md
Totktonada: Ok.

PR #56, opened 2018-02-26
Support of nesting a record into a record

### LINKED ISSUE #46: Test record within a record
- [x] generate appropriate GraphQL types;
- [x] test for selecting such objects;
- [x] test for filtering by an internal fields using `foo.bar: …` or `foo: {bar: …}` syntax.

### LINKED ISSUE #49: Avro schema with a record as a record field causes an error
Schema:
```
"user": {
    "type": "record",
    "name": "user",
    "fields": [
        {"name": "uid", "type": "long"},
        {"name": "p1", "type": "string"},
        {"name": "p2", "type": "string"},
        {
            "name": "nested",
            "type": {
                "type": "record",
                "name": "nested",
                "fields": [
                    {"name": "x", "type": "long"},
                    {"name": "y", "type": "long"}
                ]
            }
          }

    ]
}
```

Error:
`Encountered multiple types named "nested"`

Totktonada:
* Related to #46.
* Fixes #49.

### DIFF
--- a/graphql/tarantool_graphql.lua
+++ b/graphql/tarantool_graphql.lua
@@ -105,9 +105,7 @@ end
 
 --- Non-recursive version of the @{gql_type} function that returns
 --- InputObject instead of Object.
-local function gql_argument_type(state, avro_schema)
-    assert(type(state) == 'table',
-        'state must be a table, got ' .. type(state))
+local function gql_argument_type(avro_schema)
     assert(avro_schema ~= nil,
         'avro_schema must not be nil')
 
@@ -149,27 +147,53 @@ local function gql_argument_type(state, avro_schema)
     end
 end
 
-local function convert_record_fields_to_args(state, fields)
+--- Convert each field of an avro-schema to a scalar graphql type or an input
+--- object.
+---
+--- @tparam table fields list of fields of the avro-schema record fields format
+---
+--- @tparam[opt] table opts optional options:
+---
+--- * `skip_compound` -- do not add fields of record type to the arguments;
+--- default: false.
+---
+--- @treturn table `args` -- map with type names as keys and graphql types as
+--- values
+local function convert_record_fields_to_args(fields, opts)
+    assert(type(fields) == 'table',
+        'fields must be a table, got ' .. type(fields))
+
+    local opts = opts or {}
+    assert(type(opts) == 'table',
+        'opts must be a table, got ' .. type(opts))
+
+    local skip_compound = opts.skip_compound or false
+    assert(type(skip_compound) == 'boolean',
+        'skip_compound must be a boolean, got ' .. type(skip_compound))
+
     local args = {}
     for _, field in ipairs(fields) do
         assert(type(field.name) == 'string',
             ('field.name must be a string, got %s (schema %s)')
             :format(type(field.name), json.encode(field)))
-        local gql_class = gql_argument_type(state, field.type)
-        args[field.name] = nullable(gql_class)
+        if not skip_compound or avro_type(field.type) ~= 'record' then
+            local gql_class = gql_argument_type(field.type)
+            args[field.name] = nullable(gql_class)
+        end
     end
     return args
 end
 
---- Convert each field of an avro-schema to a graphql type and corresponding
---- argument for an upper graphql type.
+--- Convert each field of an avro-schema to a graphql type.
 ---
 --- @tparam table state for read state.accessor and previously filled
 --- state.types
 --- @tparam table fields fields part from an avro-schema
+---
+--- @treturn table `res` -- map with type names as keys and graphql types as
+--- values
 local function convert_record_fields(state, fields)
     local res = {}
-    local object_args = {}
     for _, field in ipairs(fields) do
         assert(type(field.name) == 'string',
             ('field.name must be a string, got %s (schema %s)')
@@ -178,9 +202,8 @@ local function convert_record_fields(state, fields)
             name = field.name,
             kind = gql_type(state, field.type),
         }
-        object_args[field.name] = nullable(res[field.name].kind)
     end
-    return res, object_args
+    return res
 end
 
 --- The function recursively converts passed avro-schema to a graphql type.
@@ -228,7 +251,7 @@ gql_type = function(state, avro_schema, collection, collection_name)
             ('avro_schema.fields must be a table, got %s (avro_schema %s)')
             :format(type(avro_schema.fields), json.encode(avro_schema)))
 
-        local fields, _ = convert_record_fields(state, avro_schema.fields)
+        local fields = convert_record_fields(state, avro_schema.fields)
 
         for _, c in ipairs((collection or {}).connections or {}) do
             assert(type(c.type) == 'string',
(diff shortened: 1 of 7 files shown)


graphql/tarantool_graphql.lua
```diff
@@ -149,27 +149,30 @@ local function gql_argument_type(state, avro_schema)
...
 
---- Convert each field of an avro-schema to a graphql type and corresponding
---- argument for an upper graphql type.
+--- Convert each field of an avro-schema to a graphql type
 ---
 --- @tparam table state for read state.accessor and previously filled
 --- state.types
 --- @tparam table fields fields part from an avro-schema
+---
+--- @treturn table res XXX
```
Khatskevich: Does 'XXX' mean obvious? Or it is TODO?
In case it is TODO, I would like to see there 'todo' instead of 'xxx', because first option do not require any explanations.

test/common/lua/multirunner.lua
```diff
@@ -12,9 +14,10 @@ local function init_shard(test_run, servers, config)
     local shard = require('shard')
     local suite = "common"
     test_run:create_cluster(servers, suite)
-    box.once('init_shard_module', function()
+    if not initialized then
         shard.init(config)
-    end)
+        initialized = true
+    end
```
Khatskevich, line 26: Can you explain, please?

test/common/lua/test_data_nested_record.lua
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }]]),
+    collections = json.decode([[{
+        "user": {
+            "schema_name": "user",
+            "connections": []
+        }
+    }]]),
+    service_fields = {
+        user = {},
```
Khatskevich, line 41: what does `user = {},` mean? Is it neccessary?

graphql/tarantool_graphql.lua
```diff
@@ -149,27 +149,30 @@ local function gql_argument_type(state, avro_schema)
     end
 end
 
-local function convert_record_fields_to_args(state, fields)
+local function convert_record_fields_to_args(state, fields, skip_compound)
```
Khatskevich: I prefer `skip_compound` -> `skip_nested`

graphql/tarantool_graphql.lua
Totktonada: XXX is like TODO, but is not commitment. It means 'smth is not good here'. When I add that to the ldoc comment it means 'I’ll fix it before considering the PR as ready'. XXX is standard mark for such things.

```
$ echo $((`ag -c '\bXXX\b' /usr/src/linux | cut -d: -f2 | tr '\n' '+'; echo 0`))
3182
$ echo $((`ag -c '\bTODO\b' /usr/src/linux | cut -d: -f2 | tr '\n' '+'; echo 0`))
4703
```

test/common/lua/test_data_nested_record.lua
Totktonada: Empty list of service fields. Don’t sure we allow to skip it, but that has sense, yep.

graphql/tarantool_graphql.lua
Totktonada: Scalar fields are nested too.

test/common/lua/multirunner.lua
Totktonada: 1. I added the 2nd test. When the default (master) instance stops (the suite has `core = app`) and starts from a snap/xlog from the previous test, the box.once will not invoke their function, so shard become uninitialized.
2. Not it works, but is not correct. We avoid the another problem: the shard module cannot be initialized twice, so we omit the 2nd initialization and proceed twice with the same (the first one) configuration.

Totktonada: I updated the code with the comment.

Totktonada: Filed the following issue for this problem: https://github.com/tarantool/graphql/issues/57

graphql/tarantool_graphql.lua
Totktonada: Fixed.

graphql/tarantool_graphql.lua
```diff
@@ -149,27 +147,53 @@ local function gql_argument_type(state, avro_schema)
     end
 end
 
-local function convert_record_fields_to_args(state, fields)
+--- Convert each field of an avro-schema to a scalar graphql type or an input
+--- object.
+---
+--- @tparam table fields list of fields the avro-schema record fields format
```
Khatskevich: I believe you forgot "in"

graphql/tarantool_graphql.lua
```diff
@@ -149,27 +147,53 @@ local function gql_argument_type(state, avro_schema)
...
+---
+--- @tparam[opt] table opts optional options:
+---
+--- * skip_compound -- do not add fields of record type to the arguments;
```
Khatskevich: I try to wrap things like `skip_compound` into back quotes. However I im not sure is it is a good practice

graphql/tarantool_graphql.lua
```diff
@@ -149,27 +147,53 @@ local function gql_argument_type(state, avro_schema)
...
+--- default: false.
+---
+--- @treturn table `args` -- map with type names as keys and graphql types as
```
Khatskevich: Why did you wrap args in ` while in other similar cases you do not to this?

graphql/tarantool_graphql.lua
```diff
@@ -149,27 +147,53 @@ local function gql_argument_type(state, avro_schema)
...
     local args = {}
     for _, field in ipairs(fields) do
         assert(type(field.name) == 'string',
             ('field.name must be a string, got %s (schema %s)')
             :format(type(field.name), json.encode(field)))
-        local gql_class = gql_argument_type(state, field.type)
-        args[field.name] = nullable(gql_class)
+        if not skip_compound or avro_type(field.type) ~= 'record' then
+            local gql_class = gql_argument_type(field.type)
+            args[field.name] = nullable(gql_class)
```
Khatskevich, line 181: Why do you call nullable without any conditions?

graphql/tarantool_graphql.lua
Totktonada: I will add 'of'.

graphql/tarantool_graphql.lua
Totktonada: I like such visual accents too.

graphql/tarantool_graphql.lua
Totktonada: `@tparam` marks the parameter name monotyped automatically, but `@treturn` is not.

graphql/tarantool_graphql.lua
Totktonada: The idea was that you can omit any argument. But AFAIR it is so even if you don’t use 'nullable'.

Khatskevich (approved):

test/common/lua/test_data_nested_record.lua
Totktonada: Skipped for now.

PR #58, opened 2018-02-28
Kh/to avro

Khatskevich:


test/common/lua/test_data_user_order.lua
```diff
@@ -112,26 +224,243 @@ function testdata.init_spaces()
         {type = 'tree', unique = false, parts = {
             O_USER_ID_FN, 'unsigned'
         }})
+    -- item_collection fields
+    local I_ITEM_ID_FN = 1
+    box.schema.create_space('item_collection')
```
Khatskevich: Forgot to delete them.

Khatskevich: Fixed.

test/common/lua/test_data_user_order.lua
```diff
@@ -41,23 +84,74 @@ testdata.meta = {
         "order_collection": {
             "schema_name": "order",
             "connections": [
+                {
+                    "type": "1:N",
+                    "name": "order__order_item",
```
Totktonada, line 90: Having schema/collection name as part of connection makes understanding of a query harder.

test/common/lua/test_data_user_order.lua
```diff
@@ -1,5 +1,27 @@
...
+-- This module helps to write tests providing simple shop database.
+--
+-- Data layout
+--
+-- models:
+--  - user: Users of the shop.
+--    - id
+--    - first_name
+--    - last_name
+--  - order: Orders of users.
```
Totktonada: I propose to clarify (comment here) that order is a shopping cart to better understanding the example.

test/common/lua/test_data_user_order.lua
```diff
@@ -112,26 +224,245 @@ function testdata.init_spaces()
...
 
+local items
 
 function testdata.fill_test_data(virtbox)
     local order_id = 1
-    for i = 1, 15 do
+    local item_id_max = #items
+    for _, item in ipairs(items) do
+       virtbox.item_collection:replace(
+               {item.id, item.description, item.name, item.price}
```
Totktonada: Broken indent.

test/common/lua/test_data_user_order.lua
```diff
@@ -112,26 +224,245 @@ function testdata.init_spaces()
...
             virtbox.order_collection:replace(
-                {order_id, i, 'order of user ' .. i})
+                { order_id, user_id, 'order of user ' .. user_id })
             order_id = order_id + 1
+            local items_cnt = 3
+            for k = 1, items_cnt do
+                order_item_cnt = order_item_cnt + 1
+                local item_id = order_item_cnt % item_id_max + 1
+                virtbox.order_item_collection:replace(
+                        {
```
Totktonada: Broken indent.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        return result
+        --return collection_to_avro(fieldType, result, subSelections, context)
+    elseif fieldTypeName == 'Interface' or fieldTypeName == 'Union' then
+        local objectType = fieldType.resolveType(result)
+        result.type = collection_to_avro(objectType, result,
+                                         subSelections, context)
+        return result
+    end
+    error('Unknown type "' .. fieldTypeName .. '" for field "' ..
+            field.name .. '"')
```
Totktonada: luacheck output:

graphql/query_to_avro.lua:62:13: accessing undefined variable field

Totktonada: FYI: our lua coding guideline suggests to avoid many concatenation operators in favor of string.format.

graphql/core/query_util.lua
```diff
@@ -0,0 +1,145 @@
...
+          collectFields(objectType, fragment.selectionSet.selections, visitedFragments, result, context)
+        end
+      end
+    end
+  end
+
+  return result
+end
+
+
```
Totktonada: Inconsistent empty line padding: two lines here btw functions, but one line in other places.

graphql/core/query_util.lua
```diff
@@ -0,0 +1,145 @@
...
+      for j = 1, #selectionSet.selections do
+        table.insert(selections, selectionSet.selections[j])
+      end
+    end
+  end
+
+  return selections
+end
+
+
```
Totktonada: Two empty lines.

graphql/core/execute.lua
```diff
@@ -230,21 +152,24 @@ local function getFieldEntry(objectType, object, fields, context)
...
     return getFieldEntry(objectType, object, fields, context)
   end)
 end
 
 return function(schema, tree, rootValue, variables, operationName)
-  local context = buildContext(schema, tree, rootValue, variables, operationName)
+  local context = query_util.buildContext(schema, tree, rootValue, variables, operationName)
+  -- The field is passed to resolve function within info attribute.
+  -- Can be used to store any data within one query.
+  context.qcontext = {}
```
Totktonada, line 172: Why you moved it from buildContext? The `qcontext` field looks as part of context to build up for me.

graphql/core/query_util.lua
```diff
@@ -0,0 +1,145 @@
+local path = (...):gsub('%.[^%.]+$', '')
```
Totktonada, line 1: It seems that you more or less investigated this functions. It would be glab to see brief comments for them.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
+--- Module for convertion GraphQl query to Avro schema.
```
Totktonada: GraphQl → GraphQL.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+--- * The best way to use this module is to just call `avro_schema` methon on
+---   compiled query object.
+local path = "graphql.core"
+local introspection = require(path .. '.introspection')
+local query_util = require(path .. '.query_util')
+
+-- module functions
+local query_to_avro = {}
+
+-- use long types, to avoid overflow problems
```
Totktonada: That it is up to designer of an original schema (from which graphql schema was deduced), we must not just upcasting ints to longs in random places of the code.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        result.type = type
+        return result
+    end
+
+    if fieldTypeName == 'Scalar' or fieldTypeName == 'Enum' then
+        return fieldType.serialize(result)
+    end
+
+    if fieldTypeName == 'Object' then
+        --result.name = "lol"
```
Totktonada: I think you forgot to remove that.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        result.type = collection_to_avro(fieldType, result,
+                                         subSelections, context)
+        return result
+        --return collection_to_avro(fieldType, result, subSelections, context)
```
Totktonada: Forgotten to remove or saved here for some purpose? If so, briefly comment why, please.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        -- not scalar
+        return complete_field_to_avro(innerType, result, subSelections, context)
+    end
+    -- scalar type
+    result.type = gql_scalar_to_avro(innerType)
+    result.type = nullable and result.type .. "*" or result.type
+    return result
+end
+
+collection_to_avro = function(objectType, object, selections, context)
```
Totktonada: I think we should not follow bad practice of graphql-lua of not commenting our own functions. You know, it makes reading by newcomers feels like nightmare.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+    local groupedFieldSet = query_util.collectFields(objectType, selections,
+                                                     {}, {}, context)
+    local result= {
```
Totktonada: Forgotten space.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        name = objectType.name,
+        fields = {}
+    }
+    if #context.namespace_parts ~= 0 then
+        result.namespace = table.concat(context.namespace_parts, ".")
+    end
+    table.insert(context.namespace_parts, result.name)
+    for _, fields in pairs(groupedFieldSet) do
+        table.insert(result.fields,
+                     field_to_avro(objectType, object, fields, context))
```
Totktonada: Python-style indent.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+function query_to_avro.convert(query)
+    local state = query.state
+    local context = query_util.buildContext(state.schema, query.ast, {},
+                                            {}, query.operation_name)
+    -- The variable is necessary to avoid fullname interferention.
+    -- Each nested Avro record creates it's namespace.
+    context.namespace_parts = {}
+    local rootType = state.schema[context.operation.operation]
+    return collection_to_avro(rootType, {},
+                              context.operation.selectionSet.selections,
```
Totktonada: Python-style indent.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+
+--- The function creates an Avro schema for a given query.
+---
+--- @tparam table query object which avro schema should be created for
+---
+--- @treturn table `avro_schema` avro schema for any `query:execute()` result.
+function query_to_avro.convert(query)
+    local state = query.state
+    local context = query_util.buildContext(state.schema, query.ast, {},
+                                            {}, query.operation_name)
```
Totktonada: Python-style indent.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        local objectType = fieldType.resolveType(result)
+        result.type = collection_to_avro(objectType, result,
+                                         subSelections, context)
+        return result
+    end
+    error('Unknown type "' .. fieldTypeName .. '" for field "' ..
+            field.name .. '"')
+end
+
+local function field_to_avro(objectType, object, fields, context)
```
Totktonada: Maybe `fields_to_avro` (field → fields) fit better for this function name?

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+    local firstField = fields[1]
+    local fieldName = firstField.name.value
+    local fieldType = introspection.fieldMap[fieldName] or
+                        objectType.fields[fieldName]
```
Totktonada: Strange indent.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+    assert(fieldType ~= nil)
+    local subSelections = query_util.mergeSelectionSets(fields)
+    local fieldTypeName = fieldType.kind.__type
+    local innerType = fieldType.kind
+    local nullable = true
+    -- In case the field is nullable, the real type is in ofType attibute.
```
Totktonada: is nullable → is non-null?

graphql/query_to_avro.lua
Totktonada: Or no… hmm.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+    if fieldTypeName == 'NonNull' then
+        innerType = innerType.ofType
+        nullable = false
+    end
+    local result = {}
+    result.name = fieldName
+    if fieldType.resolve ~= nil then
```
Totktonada: Mayble better to lookup by our table of scalars `gql_scalar_to_avro_index`? I don’t sure probing of resolve function is safe approach.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+
+local function field_to_avro(objectType, object, fields, context)
+    local firstField = fields[1]
+    local fieldName = firstField.name.value
+    local fieldType = introspection.fieldMap[fieldName] or
+                        objectType.fields[fieldName]
+    assert(fieldType ~= nil)
+    local subSelections = query_util.mergeSelectionSets(fields)
+    local fieldTypeName = fieldType.kind.__type
+    local innerType = fieldType.kind
```
Totktonada: Field type is separate type in graphql-lua AST and differs from the type of the field value, yep?

test/extra/to_avro.test.lua
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env tarantool
```
Totktonada: Proposed to add this suite into luacheck (make lint) check.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        result.type = collection_to_avro(objectType, result,
+                                         subSelections, context)
+        return result
+    end
+    error('Unknown type "' .. fieldTypeName .. '" for field "' ..
+            field.name .. '"')
+end
+
+local function field_to_avro(objectType, object, fields, context)
+    local firstField = fields[1]
```
Totktonada: We processed only one field here… what is going?

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+local function gql_scalar_to_avro(fieldType)
+    assert(fieldType.__type == "Scalar", "GraphQL scalar field expected")
+    local result = gql_scalar_to_avro_index[fieldType.name]
+    assert(result ~= nil, "Unexpected scalar type: " .. fieldType.name)
+    return result
+end
+
+local collection_to_avro
+
+local function complete_field_to_avro(fieldType, result, subSelections, context)
```
Totktonada: 'Complete' means 'set type field of the result'? I think it would look much better if we just return resulting type and use it as we need in a callee. At least we should document such unusual ways to 'return' results, when they are necessary to use.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+            type = "array",
+            items = complete_field_to_avro(innerType, {}, subSelections,
+                                           context).type
+        }
+        result.type = type
+        return result
+    end
+
+    if fieldTypeName == 'Scalar' or fieldTypeName == 'Enum' then
+        return fieldType.serialize(result)
```
Totktonada: Can you describe what is going on here? We have no any value to serialize… Also it is unexpected to see scalars here, as you process it separatelly.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    if fieldTypeName == 'Object' then
+        --result.name = "lol"
+        result.type = collection_to_avro(fieldType, result,
+                                         subSelections, context)
+        return result
+        --return collection_to_avro(fieldType, result, subSelections, context)
+    elseif fieldTypeName == 'Interface' or fieldTypeName == 'Union' then
+        local objectType = fieldType.resolveType(result)
```
Totktonada: We trying to resolve union to a certain type based on what information? We have no values here to choose one of variants. Can you describe it, please?

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        result.type = collection_to_avro(objectType, result,
+                                         subSelections, context)
```
Totktonada: Python-style indent.

graphql/query_to_avro.lua
Totktonada: Maybe 'object_to_avro' is better name? I think about a case with inner record within a collection object or with an object within an array.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,130 @@
...
+        local type = {
+            type = "array",
+            items = complete_field_to_avro(innerType, {}, subSelections,
+                                           context).type
+        }
+        result.type = type
+        return result
+    end
+
+    if fieldTypeName == 'Scalar' or fieldTypeName == 'Enum' then
```
Totktonada: I don’t know how to receive 'scalar' from our avro-schemas. Maybe we should ban it?

We have no enum / union support for now (and maybe interfaces too). I think we need to test it or stated explicitly that it need to be tested once will be available. So, I expect that you will file bunch of issues to support various types in AST or at least add notes re query schemas to existing issues for supporting of such features.

(is not for offend) If you don’t sure current implementation will work properly, maybe it is better to remove it for now to keep code compact and readable?

Totktonada (commented):
Questions and comments for the code are attached. Answers for such questions can be good basis for proper documentation and code comments.

Please, give attention to my expectation to have backlog trackerized once the feature will be implemented in the current state (see related comment).

Totktonada:
Added @SudoBobo as optional reviewer.

graphql/core/execute.lua
Khatskevich: I did it just because buildContext since now is uset in to_avro too. And it is not necessary in to_avro. Second, I suppose that we need to pass this variable from tarantool_graphql (compile or/and execute), for example to specify query timeout. What do you think?

graphql/core/query_util.lua
Khatskevich: I have just copied). I will do further investigations.

graphql/query_to_avro.lua
Khatskevich: My mistake. It should not work properly, but I did not know what to do with unions. 1. They are not implemented. 2. They are represented in different ways in gql and tarantool. A suttose at this point we should assert here.

test/extra/to_avro.test.lua
Khatskevich: Done

graphql/core/query_util.lua
Khatskevich: So, (...) stores the string passed to require when the `query_util` file was required.
gsub:
1. '%.' means just match '.'
2. '%.[^%.]+$' means match everything since the last '.'
3. the whole string just creates `path` variable which simply stores dot-delimited path to the `query_util` file relative to the file `query_util` was required from.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+    return result
+end
+
+--- Create an Avro schema for a given query.
+---
+--- @tparam table query object which avro schema should be created for
+---
+--- @treturn table `avro_schema` avro schema for any `query:execute()` result.
+function query_to_avro.convert(query)
+    local state = query.state
```
Totktonada: Optional: it is good to catch `compiled_query.avro_schema()` (period instead of colon) calls and produce a meaningfull message in the case. Just assert that the `query` parameter is a table.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+    assert(fieldType.name ~= "Map", "Map type is not supported")
+    local result = gql_scalar_to_avro_index[fieldType.name]
+    assert(result ~= nil, "Unexpected scalar type: " .. fieldType.name)
+    return result
+end
+
+-- The function converts avro type to nullable.
+-- In current tarantool/avro-schema implementation we simply add '*'
+-- to the end of type name.
+-- The function do not copy the resulting type but changes it in place.
```
Totktonada: It changes and returns changed object both. Personally I prefer to implement one of that approaches to decrease possible area to misuse a function (one can expect that it does not change the parameter, because it returns the new result).

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+--
+-- @tparam table avro schema node to be converted to nullable
+--
+-- @tresult table schema node; basically it is the passed schema node,
+-- however in nullable type implementation through unions it can be different
```
Totktonada: Don't get you in two ways:

1. As I see we always return the same table (but with change field).
2. What is `nullable type implementation through unions`? The above statement says that nulable type is implemented via '*'. Maybe more context needed here.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+    local type_type = type(avro.type)
+    if type_type == "string" then
+        avro.type = avro.type .. '*'
+        return avro
+    end
+    if type_type == "table" then
+        avro.type = make_avro_type_nullable(avro.type)
+        return avro
+    end
+    assert(false, "Avro type should be a string or table")
```
Totktonada: It is good to have `"got " .. type_type` at the end (saves some time at debug).

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+}
+local function gql_scalar_to_avro(fieldType)
+    assert(fieldType.__type == "Scalar", "GraphQL scalar field expected")
+    assert(fieldType.name ~= "Map", "Map type is not supported")
+    local result = gql_scalar_to_avro_index[fieldType.name]
+    assert(result ~= nil, "Unexpected scalar type: " .. fieldType.name)
+    return result
+end
+
+-- The function converts avro type to nullable.
```
Totktonada: The function cannot be applied to its result, yep? We need to check wheter passed type is nullable or stated non-self-applicability explicitly in the comment.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+        fieldTypeName = fieldType.__type
+    elseif NonNull ~= true then
+        -- Call complete_field second time and make result nullable.
+        result = complete_field_to_avro(fieldType, result, subSelections,
+            context, true)
+        result = make_avro_type_nullable(result)
+        return result
+    end
+
+
```
Totktonada: double newline

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+    if fieldTypeName == 'Scalar' then
+        result.type = gql_scalar_to_avro(fieldType)
+        return result
+    end
+
+    if fieldTypeName == 'Object' then
+        result.type = object_to_avro(fieldType, subSelections, context)
+        return result
+    elseif fieldTypeName == 'Interface' or fieldTypeName == 'Union' then
+        error('Interfases and Utions are not supported yet')
```
Totktonada: Interfa**s**es and U**t**ions

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+        result = make_avro_type_nullable(result)
+        return result
+    end
+
+
+    if fieldTypeName == 'List' then
+        local innerType = fieldType.ofType
+        -- Steal type from virtual object.
+        -- This is necessary because in case of arrays type should be
+        -- "completed" into results `items` field, but in case of 1:1 relations
```
Totktonada: Better to oppose an array (List) vs a record (Object) and a scalar (Scalar). That is exactly what this part of the code doing. When you say about 1:1 connection it at least yield the question what is going with nested/top-level records and scalars.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+---
+--- @tparam table selections GraphQL representations of fields which should be
+--- in the output of the query
+---
+--- @tparam table context additional information for Avro schema generation; one
+--- of the fields is `namespace_parts` -- table of names of records from the
+--- root to the current object
+---
+--- @treturn table corresponding Avro schema
+object_to_avro = function(object_type, selections, context, nullable)
```
Totktonada: Unused and undocumented parameter `nullable`.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+function query_to_avro.convert(query)
+    local state = query.state
+    local context = query_util.buildContext(state.schema, query.ast, {}, {},
+        query.operation_name)
+    -- The variable is necessary to avoid fullname interferention.
+    -- Each nested Avro record creates it's namespace.
+    context.namespace_parts = {}
+    local rootType = state.schema[context.operation.operation]
+    local selections = context.operation.selectionSet.selections
+    return object_to_avro(rootType, selections, context, false)
```
Totktonada: If you’ll remove `nullable` parameter, don’t miss to remove it here.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+    assert(false, "Avro type should be a string or table")
+end
+
+local object_to_avro
+
+local function complete_field_to_avro(fieldType, result, subSelections, context,
+        NonNull)
+    local fieldTypeName = fieldType.__type
+    if fieldTypeName == 'NonNull' then
+        -- In case the field is NonNull, the real type is in ofType attibute.
```
Totktonada: attibute → attribute

graphql/query_to_avro.lua
Totktonada: We have `--no-unused-args` in luacheck, because it is too restricting in some cases. Say, when you don’t use `self` parameter of a function.

graphql/query_to_avro.lua
```diff
@@ -0,0 +1,167 @@
...
+    end
+    assert(false, "Avro type should be a string or table")
+end
+
+local object_to_avro
+
+local function complete_field_to_avro(fieldType, result, subSelections, context,
+        NonNull)
+    local fieldTypeName = fieldType.__type
+    if fieldTypeName == 'NonNull' then
```
Totktonada: This `if` is… hmm… hard to read. Let’s look over with me:

1. The `fieldTypeName == 'NonNull'` branch.
    - Here we see renaming that works like recursion, but expressed via fall-through.
    - By the way, it cannot process several nested NonNull’s.
2. `NonNull ~= true` branch.
    - Never use `x ~= true`. Your reader does not want to normalize the code in the brain.
    - Here we use recursion and don’t use fall-through.

How to make the code readable? Options I see:

- Save `is_nonnull` or `is_nullable` flag. Process fieldType/fieldType.ofType and save the result. Make it nullable in dependence of the flag. Return the result.
- Save `is_nonnull` or `is_nullable` flag. In case of NonNull resursively call the function, but pass the flag in a parameter. Process result. Make it nullable in dependence of the flag. Return the result.

I appeal you to leave it as is if you comfortable with your approach, but I think that it is good to share argued opinions between us.

Totktonada (approved):
I think it is good enough (there was worse code from me). Despite that I leave some comments to grow our feeling of a 'healthy code'.

graphql/query_to_avro.lua
Totktonada: @Khatskevich You just ignored the comment. So I don’t know your point and I have mine preferences. I’ll fix it according to what I know — mine preferences.

graphql/query_to_avro.lua
Totktonada: @Khatskevich You just ignored the comment. I’ll fix it according to mine preferences.

graphql/query_to_avro.lua
Totktonada: Assert on nullable types is worse option then self-applicability, IMO, but ok, you fixed it.

graphql/query_to_avro.lua
Totktonada: @Khatskevich You just ignored the comment. I’ll fix it according to mine preferences.

graphql/query_to_avro.lua
Totktonada: I prefer here to leave it both return and change an input in place, because of immutability of strings in Lua.

graphql/query_to_avro.lua
Totktonada: Removed confusing part about an union.

graphql/query_to_avro.lua
Totktonada: Reworked to allow nulllable type as an input.

graphql/query_to_avro.lua
Totktonada: Reworked according to the first option. Make it returning a result using `return` instead of the mutable table parameter.

PR #62, opened 2018-03-01
Use gql_type to create top level collections

Khatskevich:
The commmit introduces the idea that the `Query` object is just a
collection. If so, it is natural to addd this type to GraphQl using
the same procedure as for other types.

This is important change, because it not only deletes current code
duplication, but allows to not repeat yourself during further coding.

### DIFF
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -206,8 +206,8 @@ end
 --- function will search through
 ---
 --- @tparam table from information about a connection bring executor to select
---- from a current collection; it is nil when the executor selecting top-level
---- objects, but has the following structure for nested collections:
+--- from a current collection; its `collection_name` is 'Query' selecting
+--- top-level objects;it has the following structure:
 ---
 ---     {
 ---         collection_name = <...> (string),
@@ -267,7 +267,7 @@ local get_index_name = function(self, collection_name, from, filter, args)
     -- If an offset is set we return it as `pivot.filter`. So, select will be
     -- performed by an index from the connection, then the result will be
     -- postprocessed using `pivot`.
-    if from ~= nil then
+    if from.collection_name ~= 'Query' then
         local connection_index =
             connection_indexes[collection_name][from.connection_name]
         local index_name = connection_index.index_name
@@ -631,8 +631,8 @@ end
 ---
 --- @tparam string collection_name name of collection to perform select
 ---
---- @tparam table from collection and connection names we arrive from/by or nil
---- as defined in the `tarantool_graphql.new` function description
+--- @tparam table from collection and connection names we arrive from/by as
+--- defined in the `tarantool_graphql.new` function description
 ---
 --- @tparam table filter subset of object fields with values by which we want
 --- to find full object(s)
@@ -685,7 +685,7 @@ local function select_internal(self, collection_name, from, filter, args, extra)
     assert(self.funcs.is_collection_exists(collection_name),
         ('cannot find collection "%s"'):format(collection_name))
     local index = self.funcs.get_index(collection_name, index_name)
-    if from ~= nil then
+    if from.collection_name ~= 'Query' then
         -- allow fullscan only for a top-level object
         assert(index ~= nil,
             ('cannot find index "%s" in space "%s"'):format(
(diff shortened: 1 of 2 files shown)


graphql/tarantool_graphql.lua
```diff
@@ -461,8 +461,7 @@ local function parse_cfg(cfg)
     local collections = table.copy(cfg.collections)
     state.collections = collections
 
-    local fields = {}
-
+    -- Prepare types which represents collections and its scalar arguments.
```
Totktonada: Strictly speaking not all possible arguments are scalar. We support non-scalar `offset` argument in case when a primary index is compound. Arguments deduced from object's fields (`object_args`) are all scalar for now.

graphql/tarantool_graphql.lua
```diff
@@ -495,42 +494,32 @@ local function parse_cfg(cfg)
...
+        name = "Query",
+        fields = {}
+    }
+    local root_collection = {
+        name = "Query",
+        connections = root_connections
+    }
+
+    local root_type = gql_type(state, root_schema, root_collection,
+                               "Query")
```
Totktonada: Personally I don’t like python indentation style like

```
long_function_name(foo,
                   bar)
```

in favor of 'just use the next indent level':

```
long_function_name(foo,
    bar)
```

My position that it is up to you, but the code looks better when an one style is used. Consider this comment as informational.

graphql/accessor_general.lua
```diff
@@ -267,7 +267,7 @@ local get_index_name = function(self, collection_name, from, filter, args)
     -- If an offset is set we return it as `pivot.filter`. So, select will be
     -- performed by an index from the connection, then the result will be
     -- postprocessed using `pivot`.
-    if from ~= nil then
+    if from.collection_name ~= "Query" then
```
Totktonada: One more note re having solid style (as usual, you can ignore that): I personally prefer to use single quotes except cases when I need to write single quote inside a string. Both are okay, but consistent using make code feels as more accurate.

graphql/tarantool_graphql.lua
```diff
@@ -495,42 +494,32 @@ local function parse_cfg(cfg)
...
+    state.schema = schema.create({
+            query = root_type
```
Totktonada: Broken indent.

graphql/tarantool_graphql.lua
```diff
@@ -495,42 +494,32 @@ local function parse_cfg(cfg)
...
-                    object_args_instance, list_args_instance, extra)
-            end,
-        }
     end
 
-    local schema = schema.create({
-        query = types.object({
-            name = 'Query',
-            fields = fields,
+    local root_connections = {}
```
Totktonada: Here terms 'connection' and 'collection' is artificial or kinda, so I propose to isolate this part of the code into a function and properly comment that this function generates avro-schema representation of GraphQL’s Query type and it exploits terms `collection` and `connection` to represent the types nesting / structure, while the corresponding fields do not belong to any real collections or connections.

graphql/accessor_general.lua
Totktonada: Please, fix `from` description in the ldoc comments to the `get_index_name`, `select_internal`, `tarantool_graphql.new` functons and fix asserts in `accessor_general.new` (disallow `nil`) inside the `__index.select` function. Or, alternatively, add check for `Query` name to the `gql_type`.

Totktonada (approved):
The change is great, thanks!

I added more or less minor comments, but the PR is generally LGTM for me.

Totktonada:
Added @SudoBobo as the reviewer to let him being on track with planned changes (the review is not required).

graphql/tarantool_graphql.lua
```diff
@@ -443,6 +443,45 @@ gql_type = function(state, avro_schema, collection, collection_name)
     end
 end
 
+--- Create virtual root collection `Query`, which has connections to any
+--- collection.
+---
+--- Actually, each GQL query starts its execution from the `Query` collection.
+--- That is why it shoult contain connections to any collection.
```
Totktonada, line 450: shoult → should

graphql/accessor_general.lua
Totktonada: > tarantool_graphql.new
> from is nil for a top-level object, otherwise it is

Totktonada: > accessor_general.lua
>     assert(from == nil or type(from) == 'table',
>         'from must be nil or a table, got ' .. type(from))

PR #65, opened 2018-03-03
Index lookup by a partial key

### LINKED ISSUE #30: Support index lookup by partial value
* [x] for connections by partial index
* [x] for top-level objects

partial mean prefix of full list of index fields

### LINKED ISSUE #38: Improve algorithm of choosing an index for top-level objects
The following cases can be handled with index lookup:

1. Handle case when fields forms a prefix of existing index parts.
2. Handle case when we can get partial (prefix) or entire index by skipping some fields.

Totktonada:
* Fixes #30.
* Fixes #38.

### DIFF
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -103,7 +103,7 @@ end
 --- Get a key to lookup index by `lookup_index_name` (part of `index_cache`).
 ---
 --- @tparam table filter filter for objects, its keys (names of fields) will
---- form the result
+--- form the result of the function
 ---
 --- @treturn string `name_list_str` (key for lookup by `lookup_index_name`)
 local function filter_names_fingerprint(filter)
@@ -123,13 +123,90 @@ local function filter_names_fingerprint(filter)
     return name_list_str
 end
 
+--- Get an index using parts tree built by @{build_index_parts_tree}.
+---
+--- @tparam table node root of the prefix tree for certain collection
+---
+--- @tparam table filter map of key-value to filter objects against
+---
+--- @treturn string `index_name` or `nil` is found index
+---
+--- @treturn number `max_branch_len` is a number of index parts will be used at
+--- lookup plus 1 (because it calculated artificial root node as well as other
+--- nodes)
+---
+--- Complexity
+--- ----------
+---
+--- In short: O(SIZE(filter)^2 * COUNT(index parts for all indexes)).
+---
+--- Say we have N fields in filter (N = SIZE(filter), M indexes and K index
+--- parts at max ((M * K) and COUNT(index parts for all indexes) both are are
+--- upside limits of nodes count in the tree). We look for successors for
+--- each filter item (<= N items) in each of the tree node (<= M * K nodes),
+--- so have O(I * N * (M * K)) of somewhat we call 'iteration' of I
+--- complexity. Most heavy operation within an iteraton is table.copy(), we
+--- can assume it has O(N) complexity. So we have overall complexity O(N^2 *
+--- (M * K)).
+---
+--- We can consider worst case scenario when any node has any of filter keys as
+--- a successor. In this case nodes count is not real constraint for recursion.
+--- In such case we can calculate complexity as iteration of weight I
+--- (calculated above as O(N^2)) and iteration count as permutations of N
+--- filter items (N!). In such case we'll have O(N^2 * N!) or O(N^(3/2) * N^N)
+--- (Stirling's approximation).
+---
+--- Expectations
+--- ------------
+---
+--- We expect typical filter size as 1 or 2 and tree depth (excluding
+--- artificial root node) of the same order. So despite horrible complexity
+--- estimation it expected to be non-so-heavy. Our guess is that it worth to
+--- try hard to find best index before a large request.
+---
+--- Future optimizations
+--- --------------------
+---
+--- * replace table.copy() with something more light: maybe 'closed set' of
+--    filter items or {remove filter[k], invoke the function, add
+---   back filter[k]} (but it needed to be done in such way that will not
+---   invalidate pairs());
+--- * cache index name btw block requests of the same collection request (when
+---   we'll have block executor) and maybe even btw different requests with the
+--    same filter keys.
+local function get_best_matched_index(node, filter)
+    local index_name = (node.index_names or {})[1]
+    local max_branch_len = 1
+
+    -- optimization: don't run the loop below if there are no successors of the
+    -- current node
+    if node.successors == nil then
+        return index_name, 1
+    end
+
+    for k, v in pairs(filter) do
+        local successor_node = (node.successors or {})[k]
+        if successor_node ~= nil then
+            local new_filter = table.copy(filter)
+            new_filter[k] = nil
+            local branch_index_name, branch_len =
+                get_best_matched_index(successor_node, new_filter)
+            branch_len = branch_len + 1
+            if branch_index_name ~= nil and branch_len > max_branch_len then
+                index_name = branch_index_name
+                max_branch_len = branch_len
+            end
+        end
+    end
+
+    return index_name, max_branch_len
+end
+
 -- XXX: raw idea: we can store field-to-field_no mapping when creating
 -- `lookup_index_name` to faster form the value_list
 
 --- Flatten filter values (transform to a list) against specific index to
 --- passing it to index:pairs().
----
---- Only full keys are supported for a compound index for now.
 ---
 --- @tparam table self the data accessor
 ---
@@ -146,6 +223,9 @@ end
 --- passed index
 ---
 --- @treturn table `value_list` the value to pass to index:pairs()
+---
+--- @treturn table `new_filter` the `filter` value w/o values extracted to
+--- `value_list`
 local function flatten_filter(self, filter, collection_name, index_name)
     assert(type(self) == 'table',
         'self must be a table, got ' .. type(self))
@@ -155,6 +235,7 @@ local function flatten_filter(self, filter, collection_name, index_name)
         'index_name must be a string, got ' .. type(index_name))
 
     local value_list = {}
+    local new_filter = table.copy(filter)
 
     -- fill value_list
     local index_meta = self.indexes[collection_name][index_name]
@@ -165,6 +246,7 @@ local function flatten_filter(self, filter, collection_name, index_name)
         local value = filter[field_name]
         if value == nil then break end
         value_list[#value_list + 1] = value
+        new_filter[field_name] = nil
     end
 
     -- check for correctness: non-empty value_list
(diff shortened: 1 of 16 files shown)


test/shard_servers/master.lua
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env tarantool
```
Khatskevich, line 1: There is a typo in commit message (tarantoo) 08529e9ff1daf416bbfc6fa4ffa99b612687b603

test/shard_servers/master.lua
```diff
@@ -1,5 +1,7 @@
...
 
+shard_initialized = false
```
Khatskevich: Can the variable be local for the module?

test/testdata/compound_index_testdata.lua
```diff
@@ -231,6 +231,13 @@ function compound_index_testdata.run_queries(gql_wrapper)
             ('RESULT\n%s'):format(yaml.encode(result)))
     end)
 
+    utils.show_trace(function()
+        local variables_1_3 = {user_str = 'user_str_b'}
+        local result = gql_query_1:execute(variables_1_3)
+        results = results .. print_and_return(
+            ('RESULT\n%s'):format(yaml.encode(result)))
```
Khatskevich: Please, If you write rests this way, at least give any unique identifiers to its output.
What I see in a result file is a lot of data and I can not find there the result of the particular query.

test/testdata/compound_index_testdata.lua
```diff
@@ -231,6 +231,13 @@ function compound_index_testdata.run_queries(gql_wrapper)
             ('RESULT\n%s'):format(yaml.encode(result)))
     end)
 
+    utils.show_trace(function()
+        local variables_1_3 = {user_str = 'user_str_b'}
```
Khatskevich: This test tests that the result is correct (user_str is user_str_b), however, does it test if the index is really used?

graphql/accessor_general.lua
```diff
@@ -121,14 +121,42 @@ local function filter_names_fingerprint(filter)
...
+    for k, v in pairs(filter) do
+        local successor_node = (node.successors or {})[k]
+        if successor_node ~= nil then
+            local new_filter = table.copy(filter)
+            new_filter[k] = nil
+            local branch_index_name, branch_len =
+                get_best_matched_index(successor_node, new_filter)
+            if branch_index_name ~= nil and branch_len > max_len then
+                index_name = branch_index_name
+                max_len = branch_len
```
Khatskevich: Should max_len = max_len + 1 be moved here?

graphql/accessor_general.lua
```diff
@@ -230,9 +248,12 @@ end
...
 --- @treturn string `index_name` is name of the found index or nil
 ---
---- @treturn table `value_list` is values list from the `filter` argument
---- ordered in the such way that can be passed to the found index (has some
---- meaning only when `index_name ~= nil`)
+--- @treturn table `new_filter` is the filter value w/o values extracted into
+--- `value_list`
+---
+--- @treturn table `value_list` (optional) is values list from the `filter`
+--- argument ordered in the such way that can be passed to the found index (has
```
Khatskevich: in such way that it can be passed

Khatskevich (commented):
Watched all except last two commits

graphql/accessor_general.lua
```diff
@@ -230,9 +248,12 @@ end
 ---
 --- @treturn string `index_name` is name of the found index or nil
 ---
---- @treturn table `value_list` is values list from the `filter` argument
---- ordered in the such way that can be passed to the found index (has some
---- meaning only when `index_name ~= nil`)
+--- @treturn table `new_filter` is the filter value w/o values extracted into
```
Khatskevich, line 302: I suggest name `brute_filter` (or `scan_filter`) for more semantics

graphql/accessor_general.lua
```diff
@@ -402,6 +441,34 @@ local function build_lookup_index_name(indexes)
     return lookup_index_name
 end
 
+--- Build `parts_tree` to use in @{get_index_name} for lookup best matching
+--- index.
+---
```
Khatskevich, line 497: It required from me some time to get the return value structure.
I suppose it would be great if you draw small schematic image here. e.g.
```
transformation example
i1: f1 f2 f3 f4
i2: f1 f2 f4
become
f1--- f2 -- f3 -- f4 --> i1
         `-- f4 --> i2
```

graphql/accessor_general.lua
```diff
@@ -121,14 +121,42 @@ local function filter_names_fingerprint(filter)
...
 end
 
+--- Get an index using parts tree built by @{build_index_parts_tree}.
+---
+--- @tparam table node root of the prefix tree for certain collection
+---
+--- @tparam table filter map of key-value to filter objects against
+---
+--- @treturn string `index_name` or `nil` is found index
+---
```
Khatskevich, line 133: I want some information about its running time here.
I tried to calculate it and I suppose it is something like O(CNT(prefix tree nodes)*SIZE(filter)) in worst case which is good for most use cases.

test/testdata/compound_index_testdata.lua
Khatskevich: Those tests do not check if new functions work correctly(
Tests on added feature needed.

test/testdata/compound_index_testdata.lua
```diff
@@ -14,6 +14,11 @@ local function print_and_return(...)
     return table.concat({...}, ' ') .. '\n'
 end
 
+local function format_result(name, query, variables, result)
```
Khatskevich, line 17: This feature makes review process much easier)
However to make the previous commit readable, could you please move this commit on top?)

Khatskevich: Actually, I do not want you to waste your time. By now it is not important anymore.

test/shard_servers/master.lua
Totktonada: Fixed.

test/shard_servers/master.lua
Totktonada: Thanks! Fixed.

test/testdata/compound_index_testdata.lua
Totktonada: 1. Done in the latter commit.
2. We check here that functionality was not broken after the change. Index lookup check depends on metainfo (#71), will be tracked by #72.

Totktonada: @Khatskevich Are you about that here?

test/testdata/compound_index_testdata.lua
Totktonada: Will be done in the scope of non-prio1 issues, tracked by #72.

graphql/accessor_general.lua
Totktonada: Fixed.

graphql/accessor_general.lua
Totktonada: Nope. It needed for comparing branches. I’ll rename it to `max_branch_len` to avoid confusion.

Totktonada: Also I’ll change description of `len` return argument as (matched parts + 1) and will describe the reason why.

graphql/accessor_general.lua
Totktonada: Done.

graphql/accessor_general.lua
Totktonada: Added.

graphql/accessor_general.lua
Totktonada: I generally agree that common names are bad for domain-specific things. I’ll consider `post_filter` names or smth like so when there will less pending PRs.

There are related questions / points:

1. `post_filter` (postprocessing filter) name will loose some of its meaning once we’ll support pushdown filtering to sotrage nodes.
2. There are two other things of the same (at least for now) structure: pivot.filter (slow case of offset) and determinator of union-type connections (#8).

test/testdata/compound_index_testdata.lua
Khatskevich: :+1:

graphql/accessor_general.lua
```diff
@@ -121,14 +121,86 @@ local function filter_names_fingerprint(filter)
...
+--- We expect typical filter size as 1 or 2 and tree depth (excluding
+--- artificial root node) of the same order. So despite horrible complexity
+--- estimation it expected to be non-so-heavy. Our guess is that it worth to
+--- try hard to find best index before a large request.
+---
+--- Future optimizations
+--- --------------------
+---
+--- * replace table.copy() with something more light (maybe 'closed set' of
```
Khatskevich: We can delete item before recursive call and insert it back after

graphql/accessor_general.lua
```diff
@@ -121,14 +121,86 @@ local function filter_names_fingerprint(filter)
...
+---
+--- Say we have N fields in filter, M indexes and K index parts at max. We look
+--- for successors for each filter item (<= N items) in each of the tree node
+--- (<= M * K nodes), so have O(I * N * (M * K)) of somewhat we call 'iteration'
+--- of I complexity. Most heavy operation within an iteraton is table.copy(),
+--- we can assume it has O(N) complexity. So we have overall complexity O(N^2 *
+--- (M * K)). Nodes count (M * K) can be limited upside as count of index
+--- parts in all indexes, so we have O(N^2 * COUNT(index parts)).
+---
+--- We can consider worst case scenario when any node has any of filter keys as
```
Khatskevich: I do not think it is useful. But it took effort and I do not insist on it's deletion.

graphql/accessor_general.lua
```diff
@@ -121,14 +121,86 @@ local function filter_names_fingerprint(filter)
...
+        return index_name, 1
+    end
+
+    for k, v in pairs(filter) do
+        local successor_node = (node.successors or {})[k]
+        if successor_node ~= nil then
+            local new_filter = table.copy(filter)
+            new_filter[k] = nil
+            local branch_index_name, branch_len =
+                get_best_matched_index(successor_node, new_filter)
```
Khatskevich, line 193: Please, move + 1 here

graphql/accessor_general.lua
```diff
@@ -121,14 +121,86 @@ local function filter_names_fingerprint(filter)
...
+---
+--- @treturn string `index_name` or `nil` is found index
+---
+--- @treturn number `len` is a number of index parts will be used at lookup
+--- plus 1 (because it calculated artificial root node as others)
+---
+--- Complexity
+--- ----------
+---
+--- In short: O(N^2 * COUNT(index parts)).
```
Khatskevich: Provide here information, that n= len(filter)

graphql/accessor_general.lua
```diff
@@ -402,6 +485,64 @@ local function build_lookup_index_name(indexes)
...
+--- Resulting table of prefix trees (contains one field for collection_1):
+---
+--- ```
+--- * collection_1:
+---  \
+---   + --> root node --> foo --> bar --> baz
+---          \             \
+---           \             + --> abc
+---            \
+---             + ------> abc --> efg --> hij
```
Khatskevich: Please draw also index names as leaf nodes

Khatskevich (approved):
Main points are negotiated verbally.
I have made minor comments, while overall it looks good enough for me.

graphql/accessor_general.lua
Totktonada: Leaf nodes are actually 'abc', 'hij' and so on, but I added index names.

graphql/accessor_general.lua
Totktonada: I suspect that it is not so easy, because of pairs() invalidation.

graphql/accessor_general.lua
Totktonada: Different views on the complexity problem: dependency of filter size in context of fixed indexes or in context of worst-case indexes. Maybe it is something one should not consider.

graphql/accessor_general.lua
Totktonada: Done. Still don’t see why it is better, but changed because of lack of any opinion about that.

graphql/accessor_general.lua
Totktonada: Done.

graphql/accessor_general.lua
Totktonada: Moved branch_len = branch_len + 1 to here.

PR #67, opened 2018-03-06
Nullable 1:1 connections

### LINKED ISSUE #44: Support nullable 1:1 connections
Consider #43 for proposed semantic.

Totktonada:
Fixes #44.

### DIFF
--- a/graphql/tarantool_graphql.lua
+++ b/graphql/tarantool_graphql.lua
@@ -230,7 +230,7 @@ end
 --- Convert each field of an avro-schema to a graphql type.
 ---
 --- @tparam table state for read state.accessor and previously filled
---- state.types
+--- state.nullable_collection_types
 --- @tparam table fields fields part from an avro-schema
 ---
 --- @treturn table `res` -- map with type names as keys and graphql types as
@@ -253,7 +253,7 @@ end
 --- The function converts passed avro-schema to a GraphQL type.
 ---
 --- @tparam table state for read state.accessor and previously filled
---- state.types (state.types are gql types)
+--- state.nullable_collection_types (those are gql types)
 --- @tparam table avro_schema input avro-schema
 --- @tparam[opt] table collection table with schema_name, connections fields
 --- described a collection (e.g. tarantool's spaces)
@@ -307,8 +307,8 @@ gql_type = function(state, avro_schema, collection, collection_name)
         for _, c in ipairs((collection or {}).connections or {}) do
             assert(type(c.type) == 'string',
                 'connection.type must be a string, got ' .. type(c.type))
-            assert(c.type == '1:1' or c.type == '1:N',
-                'connection.type must be 1:1 or 1:N, got ' .. c.type)
+            assert(c.type == '1:1' or c.type == '1:1*' or c.type == '1:N',
+                'connection.type must be 1:1, 1:1* or 1:N, got ' .. c.type)
             assert(type(c.name) == 'string',
                 'connection.name must be a string, got ' .. type(c.name))
             assert(type(c.destination_collection) == 'string',
@@ -319,16 +319,20 @@ gql_type = function(state, avro_schema, collection, collection_name)
 
             -- gql type of connection field
             local destination_type =
-                state.types[c.destination_collection]
+                state.nullable_collection_types[c.destination_collection]
             assert(destination_type ~= nil,
                 ('destination_type (named %s) must not be nil'):format(
                 c.destination_collection))
 
             local c_args
             if c.type == '1:1' then
+                destination_type = types.nonNull(destination_type)
+                c_args = state.object_arguments[c.destination_collection]
+            elseif c.type == '1:1*' then
                 c_args = state.object_arguments[c.destination_collection]
             elseif c.type == '1:N' then
-                destination_type = types.nonNull(types.list(destination_type))
+                destination_type = types.nonNull(types.list(types.nonNull(
+                    destination_type)))
                 c_args = state.all_arguments[c.destination_collection]
             else
                 error('unknown connection type: ' .. tostring(c.type))
@@ -343,6 +347,8 @@ gql_type = function(state, avro_schema, collection, collection_name)
                 resolve = function(parent, args_instance, info)
                     local destination_args_names = {}
                     local destination_args_values = {}
+                    local are_all_parts_non_null = true
+                    local are_all_parts_null = true
 
                     for _, part in ipairs(c.parts) do
                         assert(type(part.source_field) == 'string',
@@ -354,8 +360,45 @@ gql_type = function(state, avro_schema, collection, collection_name)
 
                         destination_args_names[#destination_args_names + 1] =
                             part.destination_field
+
+                        local value = parent[part.source_field]
                         destination_args_values[#destination_args_values + 1] =
-                            parent[part.source_field]
+                            value
+
+                        if value ~= nil then -- nil or box.NULL
+                            are_all_parts_null = false
+                        else
+                            are_all_parts_non_null = false
+                        end
+                    end
+
+                    -- Check FULL match constraint before request of
+                    -- destination object(s). Note that connection key parts
+                    -- can be prefix of index key parts. Zero parts count
+                    -- considered as ok by this check.
+                    local ok = are_all_parts_null or are_all_parts_non_null
+                    if not ok then -- avoid extra json.encode()
+                        assert(ok,
+                            'FULL MATCH constraint was failed: connection ' ..
+                            'key parts must be all non-nulls or all nulls; ' ..
+                            'object: ' .. json.encode(parent))
+                    end
+
+                    -- Avoid non-needed index lookup on a destination
+                    -- collection when all connection parts are null:
+                    -- * return null for 1:1* connection;
+                    -- * return {} for 1:N connection (except the case when
+                    --   source collection is the Query pseudo-collection).
+                    if collection_name ~= 'Query' and are_all_parts_null then
+                        if c.type ~= '1:1*' and c.type ~= '1:N' then
+                            -- `if` is to avoid extra json.encode
+                            assert(c.type == '1:1*' or c.type == '1:N',
+                                ('only 1:1* or 1:N connections can have ' ..
+                                'all key parts null; parent is %s from ' ..
+                                'collection "%s"'):format(json.encode(parent),
+                                tostring(collection_name)))
+                        end
+                        return c.type == '1:N' and {} or nil
                     end
 
                     local from = {
@@ -386,7 +429,10 @@ gql_type = function(state, avro_schema, collection, collection_name)
                     assert(type(objs) == 'table',
                         'objs list received from an accessor ' ..
                         'must be a table, got ' .. type(objs))
-                    if c.type == '1:1' then
+                    if c.type == '1:1' or c.type == '1:1*' then
+                        -- we expect here exactly one object even for 1:1*
+                        -- connections because we processed all-parts-are-null
+                        -- situation above
                         assert(#objs == 1,
                             'expect one matching object, got ' ..
                             tostring(#objs))
@@ -405,7 +451,7 @@ gql_type = function(state, avro_schema, collection, collection_name)
                 avro_schema.name,
             fields = fields,
         })
-        return avro_t == 'enum' and types.nonNull(res) or res
+        return avro_t == 'record' and types.nonNull(res) or res
     elseif avro_t == 'enum' then
         error('enums not implemented yet') -- XXX
     elseif avro_t == 'array' or avro_t == 'array*' then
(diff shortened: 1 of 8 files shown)


Totktonada:
@Khatskevich @SudoBobo The PR is ready for review.

graphql/tarantool_graphql.lua
```diff
@@ -307,8 +307,8 @@ gql_type = function(state, avro_schema, collection, collection_name)
         for _, c in ipairs((collection or {}).connections or {}) do
             assert(type(c.type) == 'string',
                 'connection.type must be a string, got ' .. type(c.type))
-            assert(c.type == '1:1' or c.type == '1:N',
-                'connection.type must be 1:1 or 1:N, got ' .. c.type)
+            assert(c.type == '1:1' or c.type == '1:1*' or c.type == '1:N',
+                'connection.type must be 1:1, 1:1* or 1:N, got ' .. c.type)
```
Khatskevich, line 311: I do not like * syntax. It looks like magic for the person who see it the first time

graphql/tarantool_graphql.lua
```diff
@@ -253,7 +253,7 @@ end
 --- The function converts passed avro-schema to a GraphQL type.
 ---
 --- @tparam table state for read state.accessor and previously filled
---- state.types (state.types are gql types)
+--- state.nullable_collection_types (its are gql types)
```
Khatskevich: Those are

graphql/tarantool_graphql.lua
Khatskevich: Additional field (nullable=true) would also simplify your code. Because now 1:1 and 1:1* are generally the same, but you check both

graphql/tarantool_graphql.lua
```diff
@@ -230,7 +230,7 @@ end
 --- Convert each field of an avro-schema to a graphql type.
 ---
 --- @tparam table state for read state.accessor and previously filled
---- state.types
+--- state.nullable_collection_types
```
Khatskevich, line 233: Why do you rename it?

graphql/tarantool_graphql.lua
```diff
@@ -319,16 +319,21 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
-                state.types[c.destination_collection]
+                state.nullable_collection_types[c.destination_collection]
             assert(destination_type ~= nil,
                 ('destination_type (named %s) must not be nil'):format(
                 c.destination_collection))
 
             local c_args
             if c.type == '1:1' then
+                destination_type = types.nonNull(
+                    state.nullable_collection_types[c.destination_collection])
```
Khatskevich: Why you do not reuse destination type?

Khatskevich: You can make collection nonnull 2 times? I do not understand completely what would happen in this case.

graphql/tarantool_graphql.lua
```diff
@@ -405,7 +450,7 @@ gql_type = function(state, avro_schema, collection, collection_name)
                 avro_schema.name,
             fields = fields,
         })
-        return avro_t == 'enum' and types.nonNull(res) or res
+        return avro_t == 'record' and types.nonNull(res) or res
```
Khatskevich, line 454: May be I am wrong, but I suppose that if the type is nullable or not should decide type user, but not type itself.
I suggest always return nullable types.

test/testdata/nullable_1_1_conn_testdata.lua
```diff
@@ -0,0 +1,398 @@
...
+    local function new_email(body)
+        return {
+            localpart = random_string(16):hex(),
+            domain = DOMAIN,
+            body = body,
+        }
+    end
+
+    results = results .. print_and_return(([[
+        +---------------------+
```
Khatskevich: What is that?) Why do you print it? Moreover it is not printed in shard tests. Better to just leave it as a comment?

test/testdata/nullable_1_1_conn_testdata.lua
```diff
@@ -0,0 +1,398 @@
+-- The example was inspired by [1]. Consider [2] for the problem description
```
Khatskevich, line 1: It is up to you, but I do not like those teste. In my opinion we need to:
1. Write common tests
2. Check results in tests

Pros:
1. To validate test results we do not need to read 3 result files
2. The code you copy can contain small bugs, so, it should also be read 3 times
It is just a waste of time!

Khatskevich (changes requested):
First phase, but not the last...

graphql/tarantool_graphql.lua
Totktonada: I want to be consistent with our extended avro-schema approach.

graphql/tarantool_graphql.lua
Totktonada: Fixed.

graphql/tarantool_graphql.lua
Totktonada: I generally against common words to refer specific things. This field contains only collection types (and does not contain, say, nested records) and that types converted to nullable ones (because of lazy types evaluation, consider comments). So more specific name `nullable_collection_types` is fair term for that. Maybe `collection_types_nullable` is better, though.

test/testdata/nullable_1_1_conn_testdata.lua
Totktonada: Forgot to print in shard tests. Will do. It was for convenient investigation of run of the test.

graphql/tarantool_graphql.lua
Totktonada: Interesting idea, but it would be architectural change. It is smth that we can have in our minds to consider. Thanks!

graphql/tarantool_graphql.lua
Totktonada: 1. Fixed.
2. I’m sure that collection types are nullable because it is how I save it.

test/testdata/nullable_1_1_conn_testdata.lua
Totktonada: Consider yaml module in tarantool:

third_party/lua-yaml/lyaml.cc
```
         } else if (strstr(str, "\n\n") != NULL) {
            /*  
             * Tarantool-specific: use literal style for string with empty lines.
             * Useful for tutorial().
             */
            style = YAML_LITERAL_SCALAR_STYLE;
         }   
```

I added two newlines in the begin of the string and got right output for the shard tests.

test/testdata/nullable_1_1_conn_testdata.lua
Totktonada: I think I should work on reducing code size of the tests. Maybe I should consider 'common' test suite + tap13, but I want to: run local test as just file, don’t repeat the sime initialization / running steps, working shard module reloading. I’ll consider your pain at reading in the next PRs.

graphql/tarantool_graphql.lua
```diff
@@ -354,8 +360,43 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
                         destination_args_names[#destination_args_names + 1] =
                             part.destination_field
+
+                        local value = parent[part.source_field]
                         destination_args_values[#destination_args_values + 1] =
-                            parent[part.source_field]
+                            value
+
+                        if value ~= nil then -- nil or box.NULL
+                            are_all_parts_null = false
```
Khatskevich, line 369: I suppose this code should be moved outside of the closure (as a call to a function)

graphql/tarantool_graphql.lua
```diff
@@ -354,8 +360,43 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+                    -- destination object(s)
+                    --
+                    -- note that connection key parts can be prefix of index
+                    -- key parts
+                    --
+                    -- zero parts count considered as ok by this check
+                    local is_full_match_ok = are_all_parts_null or
+                        are_all_parts_non_null
+                    if not is_full_match_ok then -- avoid extra json.encode()
+                        assert(is_full_match_ok,
```
Khatskevich: Delete if, it is useless because your assert contains the same condition

graphql/tarantool_graphql.lua
```diff
@@ -354,8 +360,43 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+                    end
+
+                    -- check FULL match constraint before request of
+                    -- destination object(s)
+                    --
+                    -- note that connection key parts can be prefix of index
+                    -- key parts
+                    --
+                    -- zero parts count considered as ok by this check
+                    local is_full_match_ok = are_all_parts_null or
```
Khatskevich: I do not like the extra variable with the strange name `is_full_match_ok`. It does not increase readability and should be used only once in the following assert, please, use just `are_all_parts_null or are_all_parts_non_null` instead (in assert).

graphql/tarantool_graphql.lua
```diff
@@ -354,8 +360,43 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+                            'object: ' .. json.encode(parent))
+                    end
+
+                    -- When all parts are null:
+                    -- * return null for 1:1* connection;
+                    -- * avoid non-needed index lookup for 1:N connection and
+                    --   return the empty list.
+                    if are_all_parts_null then
+                        assert(c.type == '1:1*' or c.type == '1:N',
+                            'only 1:1* or 1:N connections can have ' ..
```
Khatskevich: Do not split the user-visible output into lines (grep). We should follow this rule in the future too.
To maintain good code style, I propose to create a file with error messages.

Moreover, in this particular case, I believe you should provide extra information about the failed object. Because even after user sees this error, it is impossible to debug the case.

graphql/tarantool_graphql.lua
```diff
@@ -354,8 +360,43 @@ gql_type = function(state, avro_schema, collection, collection_name)
...
+
+                        if value ~= nil then -- nil or box.NULL
+                            are_all_parts_null = false
+                        else
+                            are_all_parts_non_null = false
+                        end
+                    end
+
+                    -- check FULL match constraint before request of
+                    -- destination object(s)
```
Khatskevich: Half of your comments are sentences with a dot at the end and start with a capital teller, and half of them are just text. I like the first option more. Let's follow the style?

test/testdata/nullable_1_1_conn_testdata.lua
```diff
@@ -0,0 +1,400 @@
...
+    local results = ''
+
+    local virtbox = virtbox or box.space
+
+    -- init random number generator
+    math.randomseed(PRNG_SEED)
+
+    local function new_email(body)
+        return {
+            localpart = random_string(16):hex(),
```
Khatskevich: Please, do not use random functions for pest data. This field can not be used in test output because it is not consistent.

test/testdata/nullable_1_1_conn_testdata.lua
```diff
@@ -0,0 +1,400 @@
...
+
+    utils.show_trace(function()
+        local variables_upside = {body = 'f'}
+        local result = gql_query_upside:execute(variables_upside)
+        results = results .. print_and_return(format_result(
+            'upside', query_upside, variables_upside, result))
+    end)
+
+    -- FULL MATCH constraint: connection key parts must be all non-nulls or all
+    -- nulls
```
Khatskevich: Please, create 1 failing test.

Khatskevich (changes requested):
I request some important changes.
See comments.

graphql/tarantool_graphql.lua
Totktonada: Please, read the comment for this if.

test/testdata/nullable_1_1_conn_testdata.lua
Totktonada: I have two failing tests here. Commented explicitly that they are fails.

test/testdata/nullable_1_1_conn_testdata.lua
Totktonada: I initialized PRNG with a constant seed. You can use it in the output and it will be the same.

graphql/tarantool_graphql.lua
Totktonada: Ok, I’ll use name `ok`. Don’t want to carry lines, it looks ugly here.

graphql/tarantool_graphql.lua
Totktonada: I’ll think about that before merge, but I’ll skip it now. Thanks for the point.

graphql/tarantool_graphql.lua
Totktonada: We discussed the line carrying in the case in context of tarantool code style and decided to force formal line lenght limit. So I’ll prefer to fit the limit, but I’ll try to split a message by meaningful parts.

extra info: TBD

Totktonada: extra info: added

test/testdata/nullable_1_1_conn_testdata.lua
Totktonada: Found that math.random() gives different numbers on different machines. Implemented simple PRNG in the test.

graphql/tarantool_graphql.lua
Totktonada: My approach is the following. I use small letter in the begin and omit period at the end when the comment is one clause or like so (maybe even an expanded sentence, but only one). Several sentences started from capital letter and have period at the end.

Here we have several sentences, so I reformatted this into one multi-sentence comment (capital letter + period).

graphql/tarantool_graphql.lua
Totktonada: It seems that similar effort was made by @SudoBobo in the scope of #8. I’m going to leave it as is for now. Feel free to file an issue re code refactoring if you want to track this.

Khatskevich (approved):
Everything is negotiated or fixed. ACK

Issue #73, opened 2018-03-10
Add PCRE matching support for string fields


Totktonada:
Extracted from #13.

Totktonada:
Plan:

- [x] PCRE matching for string/string* fields w/o nesting (PR #74).
- [x] PCRE matching for string/string* fields of nested records.
- [ ] PCRE matching for string/string* fields of 1:1-connected records (after #39).

Totktonada:
Sink down the issue to prio2 because of the main part of the plan (the first item) is closed.

Totktonada:
2nd bullet is done in PR #168.

Totktonada:
3rd bullet was extracted to the separate issue: #173.

PR #74, opened 2018-03-10
PCRE matching of string fields

Totktonada:
Related to #73.

### DIFF
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -8,6 +8,12 @@ local json = require('json')
 local avro_schema = require('avro_schema')
 local utils = require('graphql.utils')
 local clock = require('clock')
+local rex = utils.optional_require('rex_pcre')
+
+-- XXX: consider using [1] when it will be mature enough;
+-- look into [2] for the status.
+-- [1]: https://github.com/igormunkin/lua-re
+-- [2]: https://github.com/tarantool/tarantool/issues/2764
 
 local accessor_general = {}
 local DEF_RESULTING_OBJECT_CNT_MAX = 10000
@@ -696,6 +702,36 @@ local function validate_collections(collections, schemas)
     end
 end
 
+--- Whether an object match set of PCRE.
+---
+--- @tparam table obj an object to check
+---
+--- @tparam table pcre map with PCRE as values; names are correspond to field
+--- names of the `obj` to match
+---
+--- @treturn boolean `res` whether the `obj` object match `pcre` set of
+--- regexps.
+local function match_using_re(obj, pcre)
+    if pcre == nil then return true end
+
+    for field_name, re in pairs(pcre) do
+        -- skip an object with null in a string* field
+        if obj[field_name] == nil then
+            return false
+        end
+        assert(rex ~= nil, 'we should not pass over :compile() ' ..
+            'with a query contains PCRE matching when there are '..
+            'no lrexlib-pcre (rex_pcre) module present')
+        -- XXX: compile re once
+        local re = rex.new(re)
+        if not re:match(obj[field_name]) then
+            return false
+        end
+    end
+
+    return true
+end
+
 --- Perform unflatten, skipping, filtering, limiting of objects. This is the
 --- core of the `select_internal` function.
 ---
@@ -741,9 +777,11 @@ local function process_tuple(state, tuple, opts)
                     qstats.fetched_object_cnt, fetched_object_cnt_max))
     assert(qcontext.deadline_clock > clock.monotonic64(),
            'query execution timeout exceeded, use `timeout_ms` to increase it')
+    local collection_name = opts.collection_name
+    local pcre = opts.pcre
 
     -- convert tuple -> object
-    local obj = opts.unflatten_tuple(opts.collection_name, tuple,
+    local obj = opts.unflatten_tuple(collection_name, tuple,
         opts.default_unflatten_tuple)
 
     -- skip all items before pivot (the item pointed by offset)
@@ -755,7 +793,8 @@ local function process_tuple(state, tuple, opts)
     end
 
     -- filter out non-matching objects
-    local match = utils.is_subtable(obj, filter)
+    local match = utils.is_subtable(obj, filter) and
+        match_using_re(obj, pcre)
     if do_filter then
         if not match then return true end
     else
@@ -828,6 +867,8 @@ local function select_internal(self, collection_name, from, filter, args, extra)
     -- XXX: save type at parsing and check here
     --assert(args.offset == nil or type(args.offset) == 'number',
     --    'args.offset must be a number of nil, got ' .. type(args.offset))
+    assert(args.pcre == nil or type(args.pcre) == 'table',
+        'args.pcre must be nil or a table, got ' .. type(args.pcre))
 
     local collection = self.collections[collection_name]
     assert(collection ~= nil,
@@ -876,6 +917,7 @@ local function select_internal(self, collection_name, from, filter, args, extra)
         collection_name = collection_name,
         unflatten_tuple = self.funcs.unflatten_tuple,
         default_unflatten_tuple = default_unflatten_tuple,
+        pcre = args.pcre,
     }
 
     if index == nil then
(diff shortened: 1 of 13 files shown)


graphql/accessor_general.lua
```diff
@@ -696,6 +702,28 @@ local function validate_collections(collections, schemas)
...
+--- XXX
+local function match_using_re(obj, pcre)
+    if pcre == nil then return true end
+
+    for field_name, re in pairs(pcre) do
+        -- skip an object with null in a string* field
+        if obj[field_name] == nil then
+            return false
+        end
+        assert(rex ~= nil, 'we should not pass over :compile() ' ..
```
Khatskevich, line 722: Is this comment or assert msg?)

graphql/accessor_general.lua
```diff
@@ -696,6 +702,28 @@ local function validate_collections(collections, schemas)
     end
 end
 
+--- XXX
+local function match_using_re(obj, pcre)
+    if pcre == nil then return true end
```
Khatskevich: Is this situation normal?
I suppose assert is more suitable here and this function is called only if necessary. But it is up to you...

graphql/accessor_general.lua
```diff
@@ -741,9 +769,11 @@ local function process_tuple(state, tuple, opts)
                     qstats.fetched_object_cnt, fetched_object_cnt_max))
     assert(qcontext.deadline_clock > clock.monotonic64(),
            'query execution timeout exceeded, use `timeout_ms` to increase it')
+    local collection_name = opts.collection_name
```
Khatskevich, line 780: I do not like creating variables for one use.

graphql/accessor_general.lua
```diff
@@ -696,6 +702,28 @@ local function validate_collections(collections, schemas)
     end
 end
 
+--- XXX
+local function match_using_re(obj, pcre)
```
Khatskevich: I prefer `regexp` name over `pcre`. 
Pcre dis
1. can be confusing for newcomers
2. emphasizes implementation.

graphql/accessor_general.lua
```diff
@@ -828,6 +859,8 @@ local function select_internal(self, collection_name, from, filter, args, extra)
     -- XXX: save type at parsing and check here
     --assert(args.offset == nil or type(args.offset) == 'number',
     --    'args.offset must be a number of nil, got ' .. type(args.offset))
+    assert(args.pcre == nil or type(args.pcre) == 'table',
```
Khatskevich, line 870: I do not like, that we loose structure of filters.
I suggest:
1. Create single attribute `filter`
2. `filter` is a table which contains
    - `regexp`
    - `brute` (i do not insist on the naming, I insist on sense)
    - `index`

I believe that this approach
1. much more extensible 
2. would speed up future coding
3. do not ban such names of fields like `pcre`

graphql/accessor_general.lua
```diff
@@ -976,6 +1010,87 @@ local function init_qcontext(accessor, qcontext)
...
+local function get_primary_key_type(self, collection_name)
+    -- get name of field of primary key
+    local _, index_meta = get_primary_index_meta(
+        self, collection_name)
+
+    local offset_fields = {}
+
+    for _, field_name in ipairs(index_meta.fields) do
+        local field_type
+        local collection = self.collections[collection_name]
```
Khatskevich: Please, initialize variables which are not related to the particular loop cycle outside of the loop. It is a little coonfusing.

graphql/accessor_general.lua
```diff
@@ -976,6 +1010,87 @@ local function init_qcontext(accessor, qcontext)
...
+        local schema = self.schemas[collection.schema_name]
+        for _, field in ipairs(schema.fields) do
+            if field.name == field_name then
```
Khatskevich: What is about rested records?

graphql/accessor_general.lua
```diff
@@ -976,6 +1010,87 @@ local function init_qcontext(accessor, qcontext)
         settings.timeout_ms * 1000 * 1000
 end
 
+--- XXX
+local function get_primary_key_type(self, collection_name)
```
Khatskevich: rename pk_avro_type

graphql/accessor_general.lua
```diff
@@ -976,6 +1010,87 @@ local function init_qcontext(accessor, qcontext)
...
+end
+
+-- XXX: add string fields of a nested record / 1:1 connection
+
+--- XXX
+--- Note: it is only for list of objects and is not for destination of 1:1
+--- connection
+local function get_pcre_argument_type(self, collection_name)
+    local collection = self.collections[collection_name]
+    assert(collection ~= nil, 'cannot found collection ' ..
```
Khatskevich: Please do not write long messages to errors which should not occur. It just makes code-base huge and increases maintenance costs.

graphql/accessor_general.lua
```diff
@@ -1114,53 +1229,20 @@ function accessor_general.new(opts, funcs)
...
-                    offset_type = {
-                        name = collection_name .. '_offset',
-                        type = 'record',
-                        fields = offset_fields,
-                    }
+                -- add `pcre` argument only if lrexlib-pcre was found
+                local pcre_field
+                if rex ~= nil then
+                    local pcre_type = get_pcre_argument_type(self, collection_name)
+                    pcre_field = {name = 'pcre', type = pcre_type}
```
Khatskevich, line 1266: I have wrote above. I do not lile `pcre` name. Moreover in user-visible part.

test/testdata/compound_index_testdata.lua
```diff
@@ -459,6 +459,34 @@ function compound_index_testdata.run_queries(gql_wrapper)
...
+                    user_num: $user_num}) {
+                user_str
+                user_num
+                last_name
+                first_name
+            }
+        }
+    ]]
+
+    utils.show_trace(function()
```
Khatskevich, line 478: What is the purpose of this diff?

graphql/accessor_general.lua
Totktonada: Propose one better or I’ll leave it as is.

graphql/accessor_general.lua
Totktonada: `pcre` is like `filter`. You maybe confuse it with `rex` — module variable.

graphql/accessor_general.lua
Totktonada: I do not like unpacking some opts, but don’t unpacking others. Debatable and non-important. Going to leave as is.

graphql/accessor_general.lua
Totktonada: 1. I don’t know a developer that don’t know PCRE term.
2. It holds a standard, not an implementation. Consider comment with link to the alternative lua PCRE implementation.

I’m tentative about your proposal. Maybe.

graphql/accessor_general.lua
Totktonada: Not in the scope of this issue.

graphql/accessor_general.lua
Totktonada: Ok.

graphql/accessor_general.lua
Totktonada: XXX mark above. Will file an issue re nested records and 1:1 connections (nested filters).

graphql/accessor_general.lua
Totktonada: 1. Here all types are avro.
2. I prefer to omit abbreviations where possible.

Debatable, non-important. Going to leave as is.

graphql/accessor_general.lua
Totktonada: > cannot found collection %s

This error message is not long. You are about some other assert I guess. Propose your wording.

graphql/accessor_general.lua
Totktonada: I’m going to think about this bunch of problems in the scope of #39. Not now.

test/testdata/compound_index_testdata.lua
Totktonada: Do you read 215c1dfaa3ba7e24a83a2a3bbd3b052ba63bcc36 comment?

PR #76, opened 2018-03-13
add simple tests for @skip and @include directives

### LINKED ISSUE #45: Support directives: @skip, @include
It seems that grapql-lua supports directives already, so we need to write a test to check all works as expected. The example can be found here: http://graphql.org/learn/queries/#directives . I think the test can reuse the common_testdata dataset and conditionally include/skip a connection field. I think we need to check only a boolean variable as the expression of a directive and postpone more complex cases when it will be requested explicitly.

SudoBobo:
Closes #45

### DIFF
--- a/test/local/directives.test.lua
+++ b/test/local/directives.test.lua
@@ -1,0 +1,127 @@
+#!/usr/bin/env tarantool
+
+local fio = require('fio')
+
+-- require in-repo version of graphql/ sources despite current working directory
+package.path = fio.abspath(debug.getinfo(1).source:match("@?(.*/)")
+    :gsub('/./', '/'):gsub('/+$', '')) .. '/../../?.lua' .. ';' .. package.path
+
+local graphql = require('graphql')
+local testdata = require('test.testdata.common_testdata')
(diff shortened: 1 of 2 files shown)


test/local/directives.test.lua
```diff
@@ -0,0 +1,127 @@
...
+    local result = gql_query_1:execute(variables_1_2)
+    print(('RESULT\n%s'):format(yaml.encode(result)))
+end)
+
+local query_2 = [[
+        query user_by_order($first_name: String, $description: String, $is: Boolean) {
+            order_collection(description: $description) {
+                order_id
+                description
+                user_connection @skip(if: $is, first_name: $first_name) {
```
Totktonada: Maybe name it $skip or use not $is (we can?)?

Totktonada (approved):
Please, add `Fixes #45` to the commit message.

test/local/directives.test.lua
SudoBobo: Yes we can. It will make the test a little bit more clear

PR #82, opened 2018-03-14
add support of union connections, closes #8

### LINKED ISSUE #8: Support destination_collection deducible from a parent object
http://graphql.org/learn/schema/#union-types

Current connection format:

```
{
    name = 'connection_name_bar',
    destination_collection = 'collection_baz',
    type = '1:1',
    parts = {
        {
            source_field = 'field_name_source_1',
            destination_field = 'field_name_destination_1'
        },
        ...
    },
    index_name = 'index_name'
}   
```

Proposed the second connection format:

```
{
    name = 'connection_name_bar',
    type='1:1',
    variants = {
        {
           filter = {foo = 1, bar = 'id_1'},
           destination_collection = 'collection_baz',
           parts = {
               {   
                   source_field = 'field_name_source_1',
                   destination_field = 'field_name_destination_1'
               },  
               ... 
            }
            index_name = 'index_name'
         },  
         ... 
     }
}
```

We can move source_fields upward from variants, but I like the idea of maximum reusability of the current code (for now, at least). The format of 'filter' choosen from the same idea.

- [ ] tarantool_graphql (validation): expand connection validation with support of the second connection format;
- [ ] tarantool_graphql (graphql schema generation): for such connections: generate all possible types and construct the union as the graphql type of the corresponding connection field;
- [ ] tarantool_graphql (resolve functions): expand `from` parameter of `accessor:select()` to be list of `filter`, `from` pairs, expand collection_name to be list of such collection names (in the corresponding order);
- [ ] accessor_general: save select_internal as is, but before invoke it do the following: match `parent` with filters from the `from` argument one by one and choose the Nth collection_name from the list `collection_name`, pass the found collection name and certain `from` variant to the unchanged `select_internal`.

Debatable: the `avro_schema_changes.org` document restricts tag value type to number / string and utilizes type conversion, that seems not good for me. Maybe we must specify tag value as a value of some field, not as a key. I proposed more powerfull way that allows to reuse the existing code of our library as much as possible.

SudoBobo:
Add new type of connections - union connections, as they described in #8. With this connection type it is possible to represent a situation when entities in child table reference different parent tables. For example, let us have child table User and parent tables Basic and Premium. One user in User table may reference to a basic user info in Basic table, while another may reference to a premium info in Premium table. In GraphQL it is represented as Union type like this:
```
User {
   user_id: ID
   connection_field_name: Basic | Premium
}
```

### DIFF
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -699,14 +699,14 @@ local function validate_collections(collections, schemas)
         local schema_name = collection.schema_name
         assert(type(schema_name) == 'string',
             'collection.schema_name must be a string, got ' ..
-                type(schema_name))
+            type(schema_name))
         assert(schemas[schema_name] ~= nil,
             ('cannot find schema "%s" for collection "%s"'):format(
-            schema_name, collection_name))
+                schema_name, collection_name))
         local connections = collection.connections
         assert(connections == nil or type(connections) == 'table',
-                'collection.connections must be nil or table, got ' ..
-                type(connections))
+            'collection.connections must be nil or table, got ' ..
+            type(connections))
         for _, connection in ipairs(connections) do
             assert(type(connection) == 'table',
                 'connection must be a table, got ' .. type(connection))
@@ -715,27 +715,27 @@ local function validate_collections(collections, schemas)
                     type(connection.name))
             if connection.destination_collection then
                 assert(type(connection.destination_collection) == 'string',
-                'connection.destination_collection must be a string, got ' ..
+                    'connection.destination_collection must be a string, got ' ..
                     type(connection.destination_collection))
                 assert(type(connection.parts) == 'table',
-                'connection.parts must be a string, got ' ..
+                    'connection.parts must be a string, got ' ..
                     type(connection.parts))
                 assert(type(connection.index_name) == 'string',
-                'connection.index_name must be a string, got ' ..
+                    'connection.index_name must be a string, got ' ..
                     type(connection.index_name))
             elseif connection.variants then
                 for _, v in pairs(connection.variants) do
-                        assert(type(v.determinant) == 'table', "variant's " ..
-                            "determinant must be a table, got " ..
-                            type(v.determinant))
-                        assert(type(v.destination_collection) == 'string',
-                    'variant.destination_collection must be a string, ' ..
-                    'got ' .. type(v.destination_collection))
-                        assert(type(v.parts) == 'table',
-                    'variant.parts must be a table, got ' .. type(v.parts))
-                        assert(type(v.index_name) == 'string',
-                    'variant.index_name must be a string, got ' ..
-                        type(v.index_name))
+                    assert(type(v.determinant) == 'table', "variant's " ..
+                        "determinant must be a table, got " ..
+                        type(v.determinant))
+                    assert(type(v.destination_collection) == 'string',
+                        'variant.destination_collection must be a string, ' ..
+                        'got ' .. type(v.destination_collection))
+                    assert(type(v.parts) == 'table',
+                        'variant.parts must be a table, got ' .. type(v.parts))
+                    assert(type(v.index_name) == 'string',
+                            'variant.index_name must be a string, got ' ..
+                            type(v.index_name))
                 end
             else
                 assert(false, ('connection "%s" of collection "%s" doesn\'t ' ..
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -614,7 +614,6 @@ local function set_connection_index(c, c_name, c_type, collection_name,
     }
 end
 
-
 --- Build `connection_indexes` table (part of `index_cache`) to use in the
 --- @{get_index_name} function.
 ---
@@ -734,11 +733,11 @@ local function validate_collections(collections, schemas)
                     assert(type(v.parts) == 'table',
                         'variant.parts must be a table, got ' .. type(v.parts))
                     assert(type(v.index_name) == 'string',
-                            'variant.index_name must be a string, got ' ..
-                            type(v.index_name))
+                        'variant.index_name must be a string, got ' ..
+                        type(v.index_name))
                 end
             else
-                assert(false, ('connection "%s" of collection "%s" doesn\'t ' ..
+                assert(false, ('connection "%s" of collection "%s" does not ' ..
                     'have neither destination collection nor variants field'):
                     format(connection.name, collection_name))
             end
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -725,9 +725,9 @@ local function validate_collections(collections, schemas)
                     type(connection.index_name))
             elseif connection.variants then
                 for _, v in pairs(connection.variants) do
-                        assert(type(v.determinant) == 'table', 'variant\'s ' ..
-                    'determinant must be a table, got ' ..
-                    type(v.determinant))
+                        assert(type(v.determinant) == 'table', "variant's " ..
+                            "determinant must be a table, got " ..
+                            type(v.determinant))
                         assert(type(v.destination_collection) == 'string',
                     'variant.destination_collection must be a string, ' ..
                     'got ' .. type(v.destination_collection))
@@ -738,8 +738,9 @@ local function validate_collections(collections, schemas)
                         type(v.index_name))
                 end
             else
-                assert(false, ('collection doesn\'t have neither destination' ..
-                    'collection nor variants field'))
+                assert(false, ('connection "%s" of collection "%s" doesn\'t ' ..
+                    'have neither destination collection nor variants field'):
+                    format(connection.name, collection_name))
             end
         end
     end
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -559,6 +559,62 @@ local function build_index_parts_tree(indexes)
     return roots
 end
 
+local function set_connection_index(c, c_name, c_type, collection_name,
+                                    indexes, connection_indexes)
+    assert(type(c.index_name) == 'string',
+        'index_name must be a string, got ' .. type(c.index_name))
+
+    -- validate index_name against 'indexes'
+    local index_meta = indexes[c.destination_collection]
+    assert(type(index_meta) == 'table',
+        'index_meta must be a table, got ' .. type(index_meta))
+
+    assert(type(collection_name) == 'string', 'collection_name expected to ' ..
+        'be string, got ' .. type(collection_name))
+
+    -- validate connection parts are match or being prefix of index
+    -- fields
+    local i = 1
+    local index_fields = index_meta[c.index_name].fields
+    for _, part in ipairs(c.parts) do
+        assert(type(part.source_field) == 'string',
+            'part.source_field must be a string, got ' ..
+            type(part.source_field))
+        assert(type(part.destination_field) == 'string',
+            'part.destination_field must be a string, got ' ..
+            type(part.destination_field))
+        assert(part.destination_field == index_fields[i],
+            ('connection "%s" of collection "%s" has destination parts that ' ..
+            'is not prefix of the index "%s" parts ' ..
+            '(destination collection - "%s")'):format(c_name, collection_name,
+            c.index_name, c.destination_collection))
+        i = i + 1
+    end
+    local parts_cnt = i - 1
+
+    -- partial index of an unique index is not guaranteed to being
+    -- unique
+    assert(c_type == '1:N' or parts_cnt == #index_fields,
+        ('1:1 connection "%s" of collection "%s" ' ..
+        'has less fields than the index of "%s" collection ' ..
+        '(cannot prove uniqueness of the partial index)'):format(c_name,
+        collection_name, c.index_name, c.destination_collection))
+
+    -- validate connection type against index uniqueness (if provided)
+    if index_meta.unique ~= nil then
+        assert(c_type == '1:N' or index_meta.unique == true,
+            ('1:1 connection ("%s") cannot be implemented ' ..
+            'on top of non-unique index ("%s")'):format(
+            c_name, c.index_name))
+    end
+
+    return {
+        index_name = c.index_name,
+        connection_type = c_type,
+    }
+end
+
+
 --- Build `connection_indexes` table (part of `index_cache`) to use in the
 --- @{get_index_name} function.
 ---
@@ -581,60 +637,28 @@ local function build_connection_indexes(indexes, collections)
     assert(type(collections) == 'table', 'collections must be a table, got ' ..
         type(collections))
     local connection_indexes = {}
-    for _, collection in pairs(collections) do
+    for collection_name, collection in pairs(collections) do
         for _, c in ipairs(collection.connections) do
-            if connection_indexes[c.destination_collection] == nil then
-                connection_indexes[c.destination_collection] = {}
-            end
-            local index_name = c.index_name
-            assert(type(index_name) == 'string',
-                'index_name must be a string, got ' .. type(index_name))
+            if c.destination_collection ~= nil then
+                if connection_indexes[c.destination_collection] == nil then
+                    connection_indexes[c.destination_collection] = {}
+                end
 
-            -- validate index_name against 'indexes'
-            local index_meta = indexes[c.destination_collection]
-            assert(type(index_meta) == 'table',
-                'index_meta must be a table, got ' .. type(index_meta))
-
-            -- validate connection parts are match or being prefix of index
-            -- fields
-            local i = 1
-            local index_fields = index_meta[c.index_name].fields
-            for _, part in ipairs(c.parts) do
-                assert(type(part.source_field) == 'string',
-                    'part.source_field must be a string, got ' ..
-                    type(part.source_field))
-                assert(type(part.destination_field) == 'string',
-                    'part.destination_field must be a string, got ' ..
-                    type(part.destination_field))
-                assert(part.destination_field == index_fields[i],
-                    ('connection "%s" of collection "%s" ' ..
-                    'has destination parts that is not prefix of the index ' ..
-                    '"%s" parts'):format(c.name, c.destination_collection,
-                    c.index_name))
-                i = i + 1
-            end
-            local parts_cnt = i - 1
-
-            -- partial index of an unique index is not guaranteed to being
-            -- unique
-            assert(c.type == '1:N' or parts_cnt == #index_fields,
-                ('1:1 connection "%s" of collection "%s" ' ..
-                'has less fields than the index "%s" has (cannot prove ' ..
-                'uniqueness of the partial index)'):format(c.name,
-                c.destination_collection, c.index_name))
-
-            -- validate connection type against index uniqueness (if provided)
-            if index_meta.unique ~= nil then
-                assert(c.type == '1:N' or index_meta.unique == true,
-                    ('1:1 connection ("%s") cannot be implemented ' ..
-                    'on top of non-unique index ("%s")'):format(
-                    c.name, index_name))
+                connection_indexes[c.destination_collection][c.name] =
+                set_connection_index(c, c.name, c.type, collection_name,
+                indexes, connection_indexes)
             end
 
-            connection_indexes[c.destination_collection][c.name] = {
-                index_name = index_name,
-                connection_type = c.type,
-            }
+            if c.variants ~= nil then
+                for _, v in ipairs(c.variants) do
+                    if connection_indexes[v.destination_collection] == nil then
+                        connection_indexes[v.destination_collection] = {}
+                    end
+                    connection_indexes[v.destination_collection][c.name] =
+                    set_connection_index(v, c.name, c.type, collection_name,
+                    indexes, connection_indexes)
+                end
+            end
         end
     end
     return connection_indexes
@@ -675,29 +699,48 @@ local function validate_collections(collections, schemas)
         local schema_name = collection.schema_name
         assert(type(schema_name) == 'string',
             'collection.schema_name must be a string, got ' ..
-            type(schema_name))
+                type(schema_name))
         assert(schemas[schema_name] ~= nil,
             ('cannot find schema "%s" for collection "%s"'):format(
             schema_name, collection_name))
         local connections = collection.connections
         assert(connections == nil or type(connections) == 'table',
-            'collection.connections must be nil or table, got ' ..
-            type(connections))
+                'collection.connections must be nil or table, got ' ..
+                type(connections))
         for _, connection in ipairs(connections) do
             assert(type(connection) == 'table',
                 'connection must be a table, got ' .. type(connection))
             assert(type(connection.name) == 'string',
                 'connection.name must be a string, got ' ..
-                type(connection.name))
-            assert(type(connection.destination_collection) == 'string',
+                    type(connection.name))
+            if connection.destination_collection then
+                assert(type(connection.destination_collection) == 'string',
                 'connection.destination_collection must be a string, got ' ..
-                type(connection.destination_collection))
-            assert(type(connection.parts) == 'table',
+                    type(connection.destination_collection))
+                assert(type(connection.parts) == 'table',
                 'connection.parts must be a string, got ' ..
-                type(connection.parts))
-            assert(type(connection.index_name) == 'string',
+                    type(connection.parts))
+                assert(type(connection.index_name) == 'string',
                 'connection.index_name must be a string, got ' ..
-                type(connection.index_name))
+                    type(connection.index_name))
+            elseif connection.variants then
+                for _, v in pairs(connection.variants) do
+                        assert(type(v.determinant) == 'table', 'variant\'s ' ..
+                    'determinant must be a table, got ' ..
+                    type(v.determinant))
+                        assert(type(v.destination_collection) == 'string',
+                    'variant.destination_collection must be a string, ' ..
+                    'got ' .. type(v.destination_collection))
+                        assert(type(v.parts) == 'table',
+                    'variant.parts must be a table, got ' .. type(v.parts))
+                        assert(type(v.index_name) == 'string',
+                    'variant.index_name must be a string, got ' ..
+                        type(v.index_name))
+                end
+            else
+                assert(false, ('collection doesn\'t have neither destination' ..
+                    'collection nor variants field'))
+            end
         end
     end
 end
--- a/graphql/accessor_general.lua
+++ b/graphql/accessor_general.lua
@@ -646,7 +646,7 @@ local function build_connection_indexes(indexes, collections)
 
                 connection_indexes[c.destination_collection][c.name] =
                 set_connection_index(c, c.name, c.type, collection_name,
-                indexes, connection_indexes)
+                    indexes, connection_indexes)
             end
 
             if c.variants ~= nil then
@@ -655,8 +655,8 @@ local function build_connection_indexes(indexes, collections)
                         connection_indexes[v.destination_collection] = {}
                     end
                     connection_indexes[v.destination_collection][c.name] =
-                    set_connection_index(v, c.name, c.type, collection_name,
-                    indexes, connection_indexes)
+                        set_connection_index(v, c.name, c.type, collection_name,
+                            indexes, connection_indexes)
                 end
             end
         end
--- a/graphql/core/query_util.lua
+++ b/graphql/core/query_util.lua
@@ -74,7 +74,7 @@ function query_util.collectFields(objectType, selections, visitedFragments, resu
       end
     elseif selection.kind == 'inlineFragment' then
       if shouldIncludeNode(selection, context) and doesFragmentApply(selection, objectType, context) then
-        collectFields(objectType, selection.selectionSet.selections, visitedFragments, result, context)
+          query_util.collectFields(objectType, selection.selectionSet.selections, visitedFragments, result, context)
       end
     elseif selection.kind == 'fragmentSpread' then
       local fragmentName = selection.name.value
@@ -82,7 +82,7 @@ function query_util.collectFields(objectType, selections, visitedFragments, resu
         visitedFragments[fragmentName] = true
         local fragment = context.fragmentMap[fragmentName]
         if fragment and shouldIncludeNode(fragment, context) and doesFragmentApply(fragment, objectType, context) then
-          collectFields(objectType, fragment.selectionSet.selections, visitedFragments, result, context)
+            query_util.collectFields(objectType, fragment.selectionSet.selections, visitedFragments, result, context)
         end
       end
     end
--- a/graphql/core/query_util.lua
+++ b/graphql/core/query_util.lua
@@ -74,7 +74,7 @@ function query_util.collectFields(objectType, selections, visitedFragments, resu
       end
     elseif selection.kind == 'inlineFragment' then
       if shouldIncludeNode(selection, context) and doesFragmentApply(selection, objectType, context) then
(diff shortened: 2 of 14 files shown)


test/testdata/union_testdata.lua
```diff
@@ -0,0 +1,265 @@
...
+end
+
+function union_testdata.init_spaces()
+    -- user_collection fields
+    local U_USER_ID_FN = 2
+
+    box.once('test_space_init_spaces', function()
+        box.schema.create_space('hero_collection')
+        box.space.hero_collection:create_index('hero_id_index',
+        { type = 'tree', unique = true, parts = { U_USER_ID_FN, 'string' }}
```
Totktonada: Strictly speaking the field which number you marked via `U_USER_ID_FN` is not `user_id`. And the comment above `user_collection fields` is not relevant for hero/human/starship/hero_info. It is just small naming issue.

Side note: indent.

test/testdata/union_testdata.lua
```diff
@@ -0,0 +1,265 @@
...
+
+    shard.hero_info_collection:replace(
+    { 1827767717, 'hero_id_2'})
+end
+
+function union_testdata.drop_spaces()
+    box.space._schema:delete('oncetest_space_init_spaces')
+    box.space.human_collection:drop()
+    box.space.starship_collection:drop()
+    box.space.hero_collection:drop()
```
Totktonada, line 173: Drop `hero_info` too?

test/testdata/union_testdata.lua
```diff
@@ -0,0 +1,265 @@
...
+        "starship": {
+            "name": "starship",
+            "type": "record",
+            "fields": [
+                { "name": "hero_id", "type": "string" },
+                { "name": "model", "type": "string" },
+                { "name": "episode", "type": "string"}
+            ]
+        },
+        "hero_info": {
```
Totktonada: It seems you don’t use `hero_info` collection / connection. Is it needed, what do you think?

test/testdata/union_testdata.lua
```diff
@@ -0,0 +1,265 @@
...
+                        model
+                    }
+                }
+            }
+        }
+    ]]
+
+    utils.show_trace(function()
+        local variables_1 = {hero_id = 'hero_id_1'}
+        local gql_query_1 = gql_wrapper:compile(query)
```
Totktonada: Just note: you can compile it once and use several times.

test/testdata/union_testdata.lua
```diff
@@ -0,0 +1,265 @@
...
+        local result = gql_query_1:execute(variables_1)
+        results = results .. print_and_return(
+        ('RESULT\n%s'):format(yaml.encode(result)))
```
Totktonada: Consider using the following if you are interested (grep for the function and its use over the other tests):

```
results = results .. print_and_return(format_result(
    '1', query_1, variables_1, result))
```

Side note: indent.

graphql/accessor_general.lua
```diff
@@ -675,29 +699,48 @@ local function validate_collections(collections, schemas)
...
-                type(connection.parts))
-            assert(type(connection.index_name) == 'string',
+                    type(connection.parts))
+                assert(type(connection.index_name) == 'string',
                 'connection.index_name must be a string, got ' ..
-                type(connection.index_name))
+                    type(connection.index_name))
+            elseif connection.variants then
+                for _, v in pairs(connection.variants) do
+                        assert(type(v.determinant) == 'table', 'variant\'s ' ..
```
Totktonada: Personally I prefer to use `"variant's ..."` to better grepability of an error message. Don’t force you to follow, just shared it with you.

Side note: negative indent? :)

graphql/accessor_general.lua
```diff
@@ -675,29 +699,48 @@ local function validate_collections(collections, schemas)
...
+                    'variant.destination_collection must be a string, ' ..
+                    'got ' .. type(v.destination_collection))
+                        assert(type(v.parts) == 'table',
+                    'variant.parts must be a table, got ' .. type(v.parts))
+                        assert(type(v.index_name) == 'string',
+                    'variant.index_name must be a string, got ' ..
+                        type(v.index_name))
+                end
+            else
+                assert(false, ('collection doesn\'t have neither destination' ..
```
Totktonada: The resulting string will have the 'destinationcollection' word :)

The parenthesis around the string literal are redundant in the case, but maybe you had desire to add co{nn,ll}ection names and use `('connection "%s" of collection "%s"...'):format(...)`, then forgot?

graphql/utils.lua
```diff
@@ -12,7 +12,6 @@ function utils.show_trace(func, ...)
         function() return func(unpack(args)) end,
         function(err)
             log.info('ERROR: ' .. tostring(err))
-            log.info(debug.traceback())
```
Totktonada: Why?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+
+        local v_list_args = state.list_arguments[v.destination_collection]
+
+        union_types[#union_types + 1] = destination_type
+
+        collection_to_arguments[v.destination_collection] = v_args
+        collection_to_list_arguments[v.destination_collection] = v_list_args
+    end
+
+    local resolveType = function (result)
```
Totktonada: Don’t split `function` keyword and the following open paranthesis.

Consider '3.1) Spaces' section of [1] for motivation (in short: function call/declare vs an operator like if/while/for/…).

[1]: https://www.kernel.org/doc/Documentation/process/coding-style.rst

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                    'part.destination_field must be a string, got ' ..
+                    type(part.destination_field))
+
+                destination_args_names[#destination_args_names + 1] =
+                part.destination_field
+                local value = parent[part.source_field]
+                destination_args_values[#destination_args_values + 1] =
+                value
+
+                if value ~= nil then -- nil of box.NonNull
```
Totktonada: s/box.NonNull/box.NULL/

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                    type(part.destination_field))
+                assert(type(part.destination_field) == 'string',
+                    'part.destination_field must be a string, got ' ..
+                    type(part.destination_field))
+
+                destination_args_names[#destination_args_names + 1] =
+                part.destination_field
+                local value = parent[part.source_field]
+                destination_args_values[#destination_args_values + 1] =
+                value
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+        assert(resulting_variant, ('Variant resolving failed.'..
+            'Parent object: "%s"\n'):format(yaml.encode(parent)))
+        return resulting_variant
+    end
+
+    local field = {
+        name = c.name,
+        kind = types.union({name = c.name, types = union_types,
+                            resolveType = resolveType}),
+        arguments =  nil,
```
Totktonada: Omitting the field would lead to the same table. Why do we need to set it to `nil` explicitly?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+        end
+    end
+
+    local resolve_variant = function (parent)
+        assert(utils.do_have_keys(parent, determinant_keys),
+            ('Parent object of union object doesn\'t have determinant ' ..
+            'fields which are necessary to determine which resolving ' ..
+            'variant should be used. Union parent object:\n"%s"\n' ..
+            'Determinant keys:\n"%s"'):
+        format(yaml.encode(parent), yaml.encode(determinant_keys)))
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                    type(part.destination_field))
+                assert(type(part.destination_field) == 'string',
+                    'part.destination_field must be a string, got ' ..
+                    type(part.destination_field))
+
+                destination_args_names[#destination_args_names + 1] =
+                part.destination_field
+                local value = parent[part.source_field]
+                destination_args_values[#destination_args_values + 1] =
+                value
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
     return res
 end
 
+--- The function converts passed simple connection to a field of GraphQL type.
+--- The function converts passed avro-schema to a GraphQL type.
```
Totktonada: Forgotten copy-paste?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+--- state.nullable_collection_types (those are gql types)
+--- @tparam table avro_schema input avro-schema
+--- @tparam[opt] table collection table with schema_name, connections fields
+--- described a collection (e.g. tarantool's spaces)
+---
+--- @tparam table state for for collection types
+--- @tparam table c simple connection to create field on
+--- @tparam table collection_name name of the collection which has given
+--- connection
+local convert_simple_connection = function(state, c, collection_name)
```
Totktonada: It is better to use more meaningful names for function parameters. You can create local alias in the function body to improve code readability, like so `local c = simple_connection`. It is okay to use short names like `c` in a local scope (preferably in a short scope).

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                    object_args_instance[k] = v
+                else
+                    error(('cannot found "%s" field ("%s" value) ' ..
+                        'within allowed fields'):format(tostring(k),
+                        tostring(v)))
+                end
+            end
+            local objs = state.accessor:select(parent,
+            c.destination_collection, from,
+            object_args_instance, list_args_instance, extra)
```
Totktonada: Indent.

graphql/core/query_util.lua
```diff
@@ -74,15 +74,15 @@ function query_util.collectFields(objectType, selections, visitedFragments, resu
       end
     elseif selection.kind == 'inlineFragment' then
       if shouldIncludeNode(selection, context) and doesFragmentApply(selection, objectType, context) then
-        collectFields(objectType, selection.selectionSet.selections, visitedFragments, result, context)
+          query_util.collectFields(objectType, selection.selectionSet.selections, visitedFragments, result, context)
```
Totktonada: It is not good to use 4 spaces indent for the only two statements in the code part where 2 spaces indent is used.

graphql/tarantool_graphql.lua
```diff
@@ -8,6 +8,7 @@
 ---   passed to an accessor function in the filter argument.
 
 local json = require('json')
+local yaml = require('yaml')
```
Totktonada: I think it is better to use `json` for an error formatting to keep it on an one line (as well as for consistency with other error messages). `yaml` is much good for huge tables, but I don’t expect that tables you formatting below are huge.

graphql/tarantool_graphql.lua
Totktonada: By the way, `local function` is preferable syntax when you don’t going to use a function before its definition.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+        state.nullable_collection_types[v.destination_collection]
+        assert(destination_type ~= nil,
+            ('destination_type (named %s) must not be nil'):format(
+            v.destination_collection))
+
+        determinant_to_variant[v.determinant] = v
+
+        local v_args
+        if c.type == '1:1' then
+            --destination_type = types.nonNull(destination_type)
```
Totktonada: Two questions:

1. Why do you use nullable destination_type here? If there is a reason it need to be stated explicitly in a comment. Especially if it is some tricky implementation detail.
2. Why do leave the usage of nonNull commented? If it is something to change in the future it should be marked with XXX or TODO comment with some description.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                value
+
+                if value ~= nil then -- nil of box.NonNull
+                    are_all_parts_null = false
+                else
+                    are_all_parts_non_null = false
+                end
+            end
+
+            -- Check FULL match constraint before request of
```
Totktonada: It is the quite big code duplication. Please, consider ways to eliminate it like wrapping some meaningful part(s) to local function(s).

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+    return field
+end
+
+--- The function converts passed connection to a field of GraphQL type
+---
+--- @tparam table state for read state.accessor and previously filled
+--- state.types (state.types are gql types)
+--- @tparam table connection connection to create field on
+--- @tparam table collection_name name of the collection which have given
+--- connection
```
Totktonada, line 615: Consider adding `@return` or `@treturn`.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                return objs[1]
+            else -- c.type == '1:N'
+                return objs
+            end
+        end
+    }
+    return field
+end
+
+--- The function converts passed connection to a field of GraphQL type
```
Totktonada: AFAIR, period at the ond of short function description is a part of ldoc format (but note that `@tparam` should no have a period at the end).

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
     return res
 end
 
+--- The function converts passed simple connection to a field of GraphQL type.
+--- The function converts passed avro-schema to a GraphQL type.
+---
+--- @tparam table state for read state.accessor and previously filled
```
Totktonada: Forgotten copy-paste (the entire block)?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+local convert_simple_connection = function(state, c, collection_name)
+    assert(type(c.destination_collection) == 'string',
+        'connection.destination_collection must be a string, got ' ..
+        type(c.destination_collection))
+    assert(type(c.parts) == 'table',
+        'connection.parts must be a string, got ' .. type(c.parts))
+
+    -- gql type of connection field
+    local destination_type =
+    state.nullable_collection_types[c.destination_collection]
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                return objs
+            end
+        end,
+    }
+
+    return field
+end
+
+--- The function converts passed union connection to a field of GraphQL type.
+--- It builds connections between union collection and destination collections
```
Totktonada: > It builds connections between union collection and destination collections

1. 'union co**ll**ection'?
2. You use 'connection' term in the two different senses (common and domain-specific) in an one sentence as I understand. It is confusing a bit. You can rephrase it like so:

> It combines destination collections of passed union connection into the Union GraphQL type.

Correct me if I understood in the wrong way.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+
+    for _, v in ipairs(c.variants) do
+        assert(v.determinant, 'each variant should have a determinant')
+            assert(type(v.determinant) == 'table', 'variant\'s determinant ' ..
+            'must end be a table, got ' .. type(v.determinant))
+        assert(type(v.destination_collection) == 'string',
+            'variant.destination_collection must be a string, got ' ..
+            type(v.destination_collection))
+        assert(type(v.parts) == 'table',
+            'variant.parts must be a string, got ' .. type(v.parts))
```
Totktonada: must be a table

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+--- @tparam table state for for collection types
+--- @tparam table c simple connection to create field on
+--- @tparam table collection_name name of the collection which has given
+--- connection
+local convert_simple_connection = function(state, c, collection_name)
+    assert(type(c.destination_collection) == 'string',
+        'connection.destination_collection must be a string, got ' ..
+        type(c.destination_collection))
+    assert(type(c.parts) == 'table',
+        'connection.parts must be a string, got ' .. type(c.parts))
```
Totktonada: must be a table

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+            assert(type(v.determinant) == 'table', 'variant\'s determinant ' ..
+            'must end be a table, got ' .. type(v.determinant))
+        assert(type(v.destination_collection) == 'string',
+            'variant.destination_collection must be a string, got ' ..
+            type(v.destination_collection))
+        assert(type(v.parts) == 'table',
+            'variant.parts must be a string, got ' .. type(v.parts))
+
+        local destination_type =
+        state.nullable_collection_types[v.destination_collection]
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+end
+
+--- The function converts passed union connection to a field of GraphQL type.
+--- It builds connections between union collection and destination collections
+--- (destination collections are 'types' of a 'Union' in GraphQL).
+---
+--- @tparam table state for collection types
+--- @tparam table c union connection to create field on
+--- @tparam table collection_name name of the collection which has given
+--- connection
```
Totktonada: Consider adding of `@treturn` or `@return`.

test/testdata/union_testdata.lua
```diff
@@ -0,0 +1,265 @@
...
+
+function union_testdata.run_queries(gql_wrapper)
+    local results = ''
+
+    local query = [[
+        query obtainHeroes($hero_id: String) {
+            hero_collection(hero_id: $hero_id) {
+                hero_id
+                hero_type
+                hero_connection {
```
Totktonada, line 184: 1. Can we add common fields outside of `... on` block? If so, let’s add it to the test.
2. Are arguments supported for that field (on common fields)? If so, let’s add.
3. Are arguemnts supported on `... on` blocks? If so, let’s add.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+local convert_union_connection = function(state, c, collection_name)
+    local union_types = {}
+    local collection_to_arguments = {}
+    local collection_to_list_arguments = {}
+
+    local determinant_keys = utils.get_keys(c.variants[1].determinant)
```
Totktonada: Are we have some assert on the graphql_tarantool.new() stage that ensures c.variants is not empty?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+
+        local v_args
+        if c.type == '1:1' then
+            --destination_type = types.nonNull(destination_type)
+            v_args = state.object_arguments[v.destination_collection]
+        elseif c.type == '1:1*' then
+            v_args = state.object_arguments[v.destination_collection]
+        elseif c.type == '1:N' then
+            destination_type = types.nonNull(types.list(types.nonNull(
+            destination_type)))
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+
+    local c_args
+    if c.type == '1:1' then
+        destination_type = types.nonNull(destination_type)
+        c_args = state.object_arguments[c.destination_collection]
+    elseif c.type == '1:1*' then
+        c_args = state.object_arguments[c.destination_collection]
+    elseif c.type == '1:N' then
+        destination_type = types.nonNull(types.list(types.nonNull(
+        destination_type)))
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+
+        collection_to_arguments[v.destination_collection] = v_args
+        collection_to_list_arguments[v.destination_collection] = v_list_args
+    end
+
+    local resolveType = function (result)
+        for _, v in pairs(c.variants) do
+            local dest_collection =
+                state.nullable_collection_types[v.destination_collection]
+            if utils.do_have_keys(result, utils.get_keys(dest_collection.fields)) then
```
Totktonada: What if the field set is the same across destination collections of the union connection?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+            ('Parent object of union object doesn\'t have determinant ' ..
+            'fields which are necessary to determine which resolving ' ..
+            'variant should be used. Union parent object:\n"%s"\n' ..
+            'Determinant keys:\n"%s"'):
+        format(yaml.encode(parent), yaml.encode(determinant_keys)))
+
+        local resulting_variant
+        for determinant, variant in pairs(determinant_to_variant) do
+            local is_match = true
+            for determinant_key, determinant_value in pairs(determinant) do
```
Totktonada: Can we reuse utils.is_subtable here?

graphql/tarantool_graphql.lua
```diff
@@ -250,6 +251,387 @@ local function convert_record_fields(state, fields)
...
+                    object_args_instance[k] = v
+                else
+                    error(('cannot found "%s" field ("%s" value) ' ..
+                        'within allowed fields'):format(tostring(k),
+                        tostring(v)))
+                end
+            end
+            local objs = state.accessor:select(parent,
+            v.destination_collection, from,
+            object_args_instance, list_args_instance, extra)
```
Totktonada: Indent.

graphql/tarantool_graphql.lua
Totktonada: graphql-lua calls `resolve` (where we must determine type anyway), then `resolveType`. Why don’t reuse the resolved type in case we already have it? Consider the following PoC patch: https://gist.github.com/Totktonada/88909ec6709950ea80e7fae668259169

By the way, the patch eliminates arbitrary order of variants check because of `pairs(determinant_to_variant)` and made the code deterministic.

Totktonada: I think the code can be deduplicated in the following way:

1. Here we analize variants here and create:
    - union_types (list of destination_type);
    - list of fake single connections;
    - pass each fake single connection to the converter, create list of resolve functions.
2. In the resolve function for union connection:
    - determine variant_num (and destination_type);
    - invoke one of saved resolve functions, save result;
    - return the result and the destination_type.

Creating artificial fake single connection can be eliminated, I guess, if we’ll extract the resolve function code from the single connection converter code to the module local function.

graphql/accessor_general.lua
```diff
@@ -581,60 +637,28 @@ local function build_connection_indexes(indexes, collections)
...
-                c.destination_collection, c.index_name))
-
-            -- validate connection type against index uniqueness (if provided)
-            if index_meta.unique ~= nil then
-                assert(c.type == '1:N' or index_meta.unique == true,
-                    ('1:1 connection ("%s") cannot be implemented ' ..
-                    'on top of non-unique index ("%s")'):format(
-                    c.name, index_name))
+                connection_indexes[c.destination_collection][c.name] =
+                set_connection_index(c, c.name, c.type, collection_name,
```
Totktonada, line 647: Indent.

graphql/accessor_general.lua
```diff
@@ -581,60 +637,28 @@ local function build_connection_indexes(indexes, collections)
...
-                index_name = index_name,
-                connection_type = c.type,
-            }
+            if c.variants ~= nil then
+                for _, v in ipairs(c.variants) do
+                    if connection_indexes[v.destination_collection] == nil then
+                        connection_indexes[v.destination_collection] = {}
+                    end
+                    connection_indexes[v.destination_collection][c.name] =
+                    set_connection_index(v, c.name, c.type, collection_name,
```
Totktonada: Indent.

graphql/accessor_general.lua
```diff
@@ -581,60 +637,28 @@ local function build_connection_indexes(indexes, collections)
...
-            connection_indexes[c.destination_collection][c.name] = {
-                index_name = index_name,
-                connection_type = c.type,
-            }
+            if c.variants ~= nil then
+                for _, v in ipairs(c.variants) do
+                    if connection_indexes[v.destination_collection] == nil then
+                        connection_indexes[v.destination_collection] = {}
+                    end
+                    connection_indexes[v.destination_collection][c.name] =
```
Totktonada, line 656: It always set index for the last variant (let’s expend your test on this case). There are two option I see to mitigate this problem:

1. Pass fake 'single connections' instead of a union connection to accessor_general.
2. Expand connection_indexes to have one more level of nesting (variant_num). Expand `from` parameter to have `variant_num`.

I think the patch should try consolidate main changes either in tarantool_graphql or accessor_general, otherwise we likely spread some abstractions across both modules instead of hide it in the one. Maybe the first option is better in the scope of this view. But 'fake single connection' can be considered as redundant new abstraction, so I don’t sure.

Are you understand the question I raise in this terms? My communication skills are not good within this domain.

Totktonada (commented):
I think the implementation can be simplified in several places, the proposals are added as comments. Let’s catch me for discussions if they are needed.

graphql/tarantool_graphql.lua
Totktonada: Added the gist above as 04714dabb7759a4b4505c506640e6ebcf67ed107 on the branch sb/union_clean_fix-proposal.

test/testdata/union_testdata.lua
SudoBobo: Should we create test/utils.lua to store functions like this?

Totktonada: Good idea.

graphql/utils.lua
SudoBobo: In some cases log.info() did not work and I have to add print to see error message and stack trace. Then I deleted them inaccurately, sorry for that.

graphql/accessor_general.lua
SudoBobo: You are right, I intended to make error message with collection and connection names. Fixed.

graphql/tarantool_graphql.lua
SudoBobo: To explicitly show our intentions about union arguments?

graphql/tarantool_graphql.lua
SudoBobo: Okay, I just was not sure about things like `local c = simple_connection`. Fixed

graphql/tarantool_graphql.lua
SudoBobo: Where do you propose to make a line between 'huge' and 'regular' tables? In terms of nesting levels

graphql/tarantool_graphql.lua
Totktonada: IMHO, related comment would be better, but ok.

graphql/tarantool_graphql.lua
Totktonada: I guess that something we consider as an one object is not big. But maybe you are right in the position to don’t make such assumptions and allow an error message to be multiline. Yaml have starting `---` mark and `...` at the end (as I see, at least for objects), so borders on an error message can be easily understanded. In that case I think it is better to add newline before object dump and place it always at the end of the message. So the message will look like this:

```
some text description; object:
---
- foo: bar
...
```

SudoBobo: Okay, this message format seems good for me

graphql/tarantool_graphql.lua
SudoBobo: 1. I used nullable destination_type to fix the following graphql-lua problem: when the library tries  to check if fragment type condition is possible for given type or not it does not handle correctly case when fragment type is wrapped inside NonNull. I think it is better to add small 'unpack' from NonNull inside graphql-lua. Fixed it this way.

graphql/tarantool_graphql.lua
SudoBobo: Wrapped most part of duplicating code in functions.

test/testdata/union_testdata.lua
SudoBobo: 1. According to GraphQL spec - no. "...but unions do not define any fields, so no fields may be queried on this type without the use of typed fragments." (http://facebook.github.io/graphql/October2016/#sec-Unions)
2.
3. They are not part of GraphQL spec (though they have open issue about it in facebook/graphql) and not supported in graphql-lua

graphql/tarantool_graphql.lua
SudoBobo: We do not. Moved this to the place after assertions on existence of .determinant fields in variants (also closer to the place where determinant_keys is used)

graphql/tarantool_graphql.lua
SudoBobo: Good idea

graphql/accessor_general.lua
SudoBobo: `connection_indexes[v.destination_collection][c.name] = ...` sets indexe for a destination collection, doesn't it? As a result we will have right index for each destination collection and that's what we want.

Totktonada: It seems that I missed that that is destination collection. There is an one problem, but it is not the regression. Consider A → C and B → C connections with the same names. Only last one will be saved in the `connection_indexes` structure and A → C connection can fail during execution.

There is one more problem, but it seems to be quite synthetic: variants to the same destination collection have the same problem.

@SudoBobo Can you file an issue on that, please?

test/testdata/union_testdata.lua
Totktonada: 1. Ok. Comment and link should be added to the related code, IMO. And maybe additionally to the ldoc of tarantool_graphql module (at the top of module) to give an user info re possible corner cases.
2. Need to be commented in the same way.
3. The same, comment + link. But I think we should additionally open an inssue on that, because it reduces feature capacity of union connections in compare with 'single' ones (consider example with replacing 'single' connection with union one with an one variant).

graphql/tarantool_graphql.lua
Totktonada: Seems to be right way to fix it.

Union type is correct (nullable/non-null) and null value I hope will cause an error for non-null connection. Fragment will ignore non-null 'flag' on got type when verifying got type with needle type. It seems to be okay, because we cannot use `!` to control nullability on each fragment in place.

test/testdata/union_testdata.lua
Totktonada: You are deleted it, okay. The comment is not more actual.

graphql/accessor_general.lua
SudoBobo: @Totktonada  done #88

Totktonada (approved):
LGTM.

PR #92, opened 2018-03-21
Microbenchmarking

### LINKED ISSUE #78: Microbenchmarking
* one level microbenchmark on spaces: test index lookup speed and so on
* two level microbenchmark on shard: test network throughput utilization
* three level microbenchmark on shard with 100 matching objects on the first level, many ones on the second level and 1 on the third level

Totktonada:
Closes #78.


test/common/lua/multirunner.lua
```diff
@@ -1,15 +1,61 @@
...
+local net_box = require('net.box')
+local utils = require('graphql.utils')
 
 local function instance_uri(instance_id)
     local socket_dir = require('fio').cwd()
     local uri = ('%s/shard%s.sock'):format(socket_dir, instance_id)
     return uri
 end
 
+local SERVERS_DEFAULT = {'shard1', 'shard2', 'shard3', 'shard4'}
```
SudoBobo: SERVERS_DEFAULT and SHARD_CONF_DEFAULT.servers are a little bit confusing

test/common/lua/multirunner.lua
```diff
@@ -40,75 +86,83 @@ local function for_each_server(shard, func)
...
+    local callback = opts.callback
+
+    assert(test_run ~= nil)
+    assert(init_function ~= nil)
+    assert(cleanup_function ~= nil)
+    assert(callback ~= nil)
+
+    if conf_type == 'space' then
+        init_function()
+        callback("Local (box)", nil)
```
SudoBobo: why not `callback(conf_name, nil)`?

test/common/lua/multirunner.lua
```diff
@@ -40,75 +86,83 @@ local function for_each_server(shard, func)
...
+        local cleanup_script = string.dump(cleanup_function)
+
+        local shard = init_shard(test_run, servers, shard_conf)
+
+        for_each_server(shard, function(uri)
+            local c = net_box.connect(uri)
+            c:eval(init_script)
+        end)
+
+        callback(conf_name, shard)
```
SudoBobo: Are you sure that this function should be called `callback`? It is not called after any actions  so it should not be called like that. Maybe `command` as it is something we want from our shards to run?

test/bench/nesting-1.test.lua
```diff
@@ -0,0 +1,147 @@
...
+    for i = 2, ITERATIONS do
+        local result = yaml.encode(gql_query_1:execute(variables_1))
+        checksum:update(result .. i)
+        fiber.yield()
+    end
+
+    local end_time = clock.monotonic64()
+
+    print('Iterations: ' .. tostring(ITERATIONS))
+    print('Checksum: ' .. tostring(checksum:result()))
```
SudoBobo: What for do we need `Checksum`?

test/bench/nesting-1.test.lua
```diff
@@ -0,0 +1,147 @@
...
+    local rps = ITERATIONS / duration
+
+    return {
+        duration = duration,
+        latency = latency,
+        rps = rps,
+    }
+end
+
+local function octal(...)
```
SudoBobo: Move to test/utils?

test/bench/nesting-1.test.lua
Totktonada: Result correctness check. W/o a checksum verification we can measure something other then we expected.

test/bench/nesting-1.test.lua
Totktonada: Will use `tonumber('644', 8)`.

test/common/lua/multirunner.lua
Totktonada: @Khatskevich is the original author.

test/common/lua/multirunner.lua
Totktonada: You are right, fixed.

test/common/lua/multirunner.lua
Khatskevich: I do not insist. It seems that callback is not the best option, however, I suppose that this small detail does not deserve as much attention.

Totktonada: Will change to `workload`.

test/common/lua/multirunner.lua
Totktonada: Added comments.

test/bench/nesting-1.test.lua
Totktonada: Such approach is used in compiler benchmarks to make sure the compiler does not omit executing of 'dead' (unused) code of benchmark. So we need to print entire output or checksum. Print possibly large output can change timings significantly, so we stick to checksum approach.

Totktonada:
Split workload from benchmarking loop code.

Totktonada:
Added planned set of microbenchmarks.

Totktonada:
Increased time limit for requests in bench/*.test.lua.

SudoBobo (approved):

PR #106, opened 2018-04-04
Add missing numeric avro types and bool

### LINKED ISSUE #31: Support for float, double and boolean avro types

opomuc:
Closes #31

### DIFF
--- a/graphql/core/types.lua
+++ b/graphql/core/types.lua
@@ -205,7 +205,7 @@ end
 
 types.int = types.scalar({
   name = 'Int',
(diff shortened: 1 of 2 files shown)


Totktonada:
Questions to think about:

1. `double` field is added to the `types` module in the `tarantool_graphql` module.
2. `Float` is 64-bt in GraphQL, AFAIR. We need to convert avro-schema double into GraphQL Float. Maybe we need also create Double type that will be full alias of the Float (and can be used instead in any place). Possible problem: use a Double variable for a Float argument and vice versa.
3. You are use `node.kind == 'double'`, but core/parser.lua does not emit such nodes (it emit float ones).
4. No tests.

Totktonada:
Closed in favor of PR #128. Roman, thanks for the effort!

PR #117, opened 2018-04-11
Support avro-schema references for records

### LINKED ISSUE #116: Add support for type references by name in avro-schemas
```
{
  "name": "users",
  "type": "record",
  "fields": [
      {
        "name": "User_A",
        "type": {
	    "name": "user",
            "type": "record",
            "fields": [
              {"name": "username", "type": "string"},
              {"name": "phone", "type": "long"},
              {"name": "age", "type": "int"}
            ]
        }
      },
      {"name": "User_B", "type": "user"}
    ]
}
```

Totktonada:
Fixes #116.

### DIFF
--- a/graphql/tarantool_graphql.lua
+++ b/graphql/tarantool_graphql.lua
@@ -56,38 +56,62 @@ local default_instance
 -- forward declarations
 local gql_type
 
-local function avro_type(avro_schema)
+local function is_scalar_type(avro_schema_type)
+    check(avro_schema_type, 'avro_schema_type', 'string')
+
+    local scalar_types = {
+        ['int'] = true,
+        ['int*'] = true,
+        ['long'] = true,
+        ['long*'] = true,
+--[[
+        ['float'] = true,
+        ['float*'] = true,
+        ['double'] = true,
+        ['double*'] = true,
+        ['boolean'] = true,
+        ['boolean*'] = true,
+]]--
+        ['string'] = true,
+        ['string*'] = true,
+        ['null'] = true,
+    }
+
+    return scalar_types[avro_schema_type] or false
+end
+
+local function is_compound_type(avro_schema_type)
+    check(avro_schema_type, 'avro_schema_type', 'string')
+
+    local compound_types = {
+        ['record'] = true,
+        ['record*'] = true,
+        ['array'] = true,
+        ['array*'] = true,
+        ['map'] = true,
+        ['map*'] = true,
+    }
+
+    return compound_types[avro_schema_type] or false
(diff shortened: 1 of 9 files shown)


graphql/tarantool_graphql.lua
```diff
@@ -949,6 +970,8 @@ gql_type = function(state, avro_schema, collection, collection_name, field_name)
                 avro_schema.name,
             fields = fields,
         })
+        state.declarations[avro_schema.name] = types.nonNull(res)
```
SudoBobo: As far as I understand we do not check anywhere if names of named types in our schema are unique (in context of type defining, not referencing) Avro-schema spec states: "A schema or protocol may not contain multiple definitions of a fullname" Moreover on line 973 we may  have over writing, haven't we? As in our typical case we have quite massive schemas, unconscious type re-definition may be a problem. So my idea to add some assertion for type re-defenition.

graphql/tarantool_graphql.lua
```diff
@@ -980,6 +1003,12 @@ gql_type = function(state, avro_schema, collection, collection_name, field_name)
     elseif avro_t == 'union' then
         return create_gql_union(state, avro_schema, field_name)
     else
+        if type(avro_schema) == 'string' then
+            if state.declarations[avro_schema] ~= nil then
+                return state.declarations[avro_schema]
+            end
+        end
+
```
SudoBobo: Do I understand right that order of definitions in our avro-schema matters? As far as I understand now if we try to put type reference BEFORE type definition in our avro-schema we will have 'unrecognized avro-schema type'. Maybe it is worthy to make additional check and return something like 'reference of yet undefined avro-schema type'

graphql/tarantool_graphql.lua
Totktonada: It is definitions, sorry. Will rename. Assert seems to make sense for proper error message, will add.

graphql/tarantool_graphql.lua
Totktonada: Order is matters. You cited the standard right above :)

'unrecognized avro-schema type' seems ok for me. I’ll see whether adding additional check will looks good.

Totktonada: We fall into this error in the two cases:

1. Forward reference (forbidden by the standard).
2. Non-defined type.

I don’t want to traverse all schema to give some-order-better error message. Will leave as is.

Totktonada:
@SudoBobo Force-pushed.

SudoBobo (approved):

PR #168, opened 2018-06-06
Split-up modules and refactor arguments converting

### LINKED ISSUE #166: Improve and refactor arguments generation
That will allow to implement #163, #73, #46.

Possibly related problem: #120.

Totktonada:
Fixes #166.


graphql/convert_schema/arguments.lua
```diff
@@ -0,0 +1,162 @@
...
+local function convert(avro_schema, opts)
+    check(avro_schema, 'avro_schema', 'table', 'string')
+    check(opts, 'opts', 'table', 'nil')
+
+    local opts = opts or {}
+    local type_name = opts.type_name
+    local context = opts.context
+
+    check(type_name, 'type_name', 'string', 'nil')
+    check(context, 'context', 'table')
```
SudoBobo, line 40: Seems a little bit strange. Context is in opts, but it is mandatory argument. Why not make a separate argument? (Maybe I missed something about our concept of opts)

graphql/convert_schema/arguments.lua
```diff
@@ -0,0 +1,162 @@
...
+
+    local avro_t = avro_helpers.avro_type(avro_schema)
+
+    if avro_t == 'record' or avro_t == 'record*' then
+        if type(avro_schema.name) ~= 'string' then -- avoid extra json.encode()
+            assert(type(avro_schema.name) == 'string',
+                ('avro_schema.name must be a string, got %s (avro_schema %s)')
+                :format(type(avro_schema.name), json.encode(avro_schema)))
```
SudoBobo, line 48: Maybe remove `assert(sure_to_be_false_condition, message)` with `error(message)` ?

graphql/convert_schema/arguments.lua
```diff
@@ -0,0 +1,162 @@
...
+                ('field.name must be a string, got %s (schema %s)')
+                :format(type(field.name), json.encode(field)))
+        end
+
+        -- We preserve a type name of an uppermost InputObject that starts from
+        -- the collection name to allow use it for variables.
+        local avro_t = avro_helpers.avro_type(field.type)
+        local type_name
+        if (avro_t == 'record' or avro_t == 'record*') and
+                field.type.name:startswith(root_name) then
```
SudoBobo, line 146: Why 2 tabs instead of one?

graphql/convert_schema/arguments.lua
```diff
@@ -0,0 +1,162 @@
...
+        end
+
+        table.insert(context.path, field.name)
+        context.field_name = field.name
+        args[field.name] = convert(field.type, {
+            context = context,
+            type_name = type_name,
+        })
+        context.field_name = nil
+        table.remove(context.path, #context.path)
```
SudoBobo, line 157: Do we really need to use this mutable `context` variable? What do you think about passing `(full) path` and a `field name` strings into `convert()`? It seems to work the same way as with this `table.remove(context.path,context.path)` If I am not clear, we can discuss it verbaly

SudoBobo: It just seems a little bit overcomplicated to use explicit stack where we can just pass arguments

SudoBobo (changes requested):

graphql/convert_schema/arguments.lua
Totktonada: Maybe I’ll make it optional sooner or later, expose `convert_schema.arguments.convert` as public function and remove `arguments.convert_record_fields`.

graphql/convert_schema/arguments.lua
Totktonada: It would be good to use error / assert according to some specific rules (@Khatskevich had a proposal on the topic, AFAIR?). Now it is arbitrary with bias toward assert.

graphql/convert_schema/arguments.lua
Totktonada: To visually distinguish if-condition from if-branch-body. It seems not being most popular indent style, but I have to use it.

graphql/impl.lua
```diff
@@ -0,0 +1,297 @@
+--- Implementation of module-level functions and functions of instances of the
+--- graphql library and a compiled query.
+
```
SudoBobo, line 3: Are you sure about naming? "Implementation" sounds quite wide and uncertain, maybe something like "Instance" and think about as a set of interface functions, connecting user and inner mechanisms of tarantool graphql?

graphql/convert_schema/arguments.lua
Totktonada: My intension was hide details of the context inside a variable and use/change needed things only where it is needed. It allows to avoid formal arguments that just pass-through them to call of an another function.

graphql/impl.lua
Totktonada: Instance was the first variant I consider about :) I was confused with the fact that require('graphql.instance') would return something that is not an instance.

Now I think about it as *.c (impl.lua) and *.h (init.lua) files. Maybe I should rename it to init_impl.lua to make analogue more clear.

graphql/convert_schema/arguments.lua
SudoBobo: But it's all in if-condition, I mean, both lines

Totktonada:
```
<0 indent>if <cond start>
<2 indent><cond middle>
<2 indent><cond end> then
<1 indent><if branch body>
<0 indent>end
```

SudoBobo: Ok, got it

graphql/impl.lua
SudoBobo: Ok, got your point

SudoBobo (approved):

PR #196, opened 2018-07-06
WIP: Support q_select

### LINKED ISSUE #195: Use parallel select on shard
Now it is possible by using q_select instead of secondary_select on >=shard-2.1-3-g1e9848c. But we want to support that in secondary_select when it is run on tarantool-1.10, so secondary_select whould be the default option. Proposed to add `shard_use_q_select` option.

Totktonada:
We need to insert data to all zones when fill test data or setup
tarantool replication.

TBD: setup full mesh (master-master) replication.

Fixes #195.


Totktonada:

* Archive the repository.

PR #199, opened 2018-07-29
Use box.schema.SYSTEM_ID_{MIN,MAX}

Totktonada:


graphql/simple_config.lua
```diff
@@ -159,7 +158,7 @@ function simple_config.get_spaces_formats()
     local FORMAT = 7
     local NAME = 3
     for _, s in box.space._space:pairs() do
-        if not is_system_space(s) then
+        if not space_is_system(s) then
```
SudoBobo, line 161: Why do you prefer the new name? I am ok, just want to know your logic.

SudoBobo (approved):

graphql/simple_config.lua
Totktonada: 'Is [it a] system space' sounds like 'whether that object of auxiliary type is system space'.
'[Whether] the space is a system one' sounds 'whether that object, which is a space, is a system space'.

I don’t strong in grammar, but the second variant feels more natural for our case.

PR #202, opened 2018-08-03
Fix Travis CI cron job

Totktonada:
It fails like so:

Pushing package: tarantool-graphql-0.0.1.2-1.fc27.src.rpm
Error {"filename":["has already been taken"]}

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ deploy:
     skip_cleanup: true
     on:
       branch: master
-      condition: -n "${OS}"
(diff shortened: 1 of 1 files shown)


SudoBobo (approved):
Ok with changes. Just curious: why do we have weekly jobs?

Totktonada:
Because not all dependencies have fixed version and we want to know if something became broken even when no merges were made for a long time.

By the way, it worth to add testing on latest git versions of avro-schema and shard. I filed #203.

PR #204, opened 2018-08-09
Fix name clash b/w 1:1 connections and collections

### LINKED ISSUE #104: Connection cannot be named as destination collection in some cases
The example is `user_to_passport_c` (the `_c` postfix is mandatory) in the nesting-3-100-100-1.test.lua benchmark. It seems that we cannot avoid it while fetching type information in the executor by type name (I don’t sure, though) and we need to introduce namespaces.

### LINKED ISSUE #191: Name clash when 1:1 connection and multi-head 1:1 connection variant have the same name
Object (I guess for the variant) and InputObject for filtering by a connection have the same names.

1. We can use bare names only for box objects, but not for the variant type itself, I think (including muti-head connection).
2. We should not use a bare name for an InputObject for filtering by a connection.

SudoBobo:
Before we could not use the same name for collection and 1:1 connection (or collection and 1:1 multi-head connection variant) because this created two different types with same names in GraphQL schema: Object type for collection and InputObject type for connection. Now we can use the same names, since we add 'result_' to the name of GraphQL collection Object type.

Close #191, close #104

### DIFF
--- a/graphql/convert_schema/types.lua
+++ b/graphql/convert_schema/types.lua
@@ -373,7 +373,7 @@ function types.convert(state, avro_schema, opts)
     if context == nil then
         context = {
             field_name = nil,
-            path = {},
+            path = {'results'},
         }
     end
 
--- a/test/extra/connections_1_1_name_clash.test.lua
+++ b/test/extra/connections_1_1_name_clash.test.lua
@@ -1,0 +1,350 @@
+#!/usr/bin/env tarantool
+local fio = require('fio')
+local json = require('json')
+local test = require('tap').test('connections 1:1 name clash')
+
+-- require in-repo version of graphql/ sources despite current working directory
+package.path = fio.abspath(debug.getinfo(1).source:match("@?(.*/)")
+    :gsub('/./', '/'):gsub('/+$', '')) .. '/../../?.lua' .. ';' ..
+    package.path
+
+local graphql = require('graphql')
+
+box.cfg{ wal_mode="none" }
+test:plan(3)
+
+local schemas = json.decode([[{
+    "hero": {
+        "name": "hero",
+        "type": "record",
+        "fields": [
+            { "name": "hero_id", "type": "string" },
+            { "name": "hero_type", "type" : "string" }
+        ]
+    },
+    "human": {
+        "name": "human",
+        "type": "record",
+        "fields": [
+            { "name": "hero_id", "type": "string" }
+        ]
+    },
+    "starship": {
+        "name": "starship",
+        "type": "record",
+        "fields": [
+            { "name": "hero_id", "type": "string" }
+        ]
+    },
+    "hero_info": {
+        "name": "hero_info",
+        "type": "record",
+        "fields": [
+            { "name": "hero_id", "type": "string" }
+        ]
+    },
+    "hero_meta": {
+        "name": "hero_meta",
+        "type": "record",
+        "fields": [
+            { "name": "hero_id", "type": "string" }
+        ]
+    }
+}]])
+
+-- In the following tests we use different collections with different
+-- potential name clash cases. We check that such combinations do not cause
+-- name clash errors.
+
+-- Potential clash in the second connection ("hero_info_collection" connection).
+-- Connection name has the same name as it's destination_collection.
+local collections_1 = json.decode([[{
+    "hero_collection": {
+        "schema_name": "hero",
+        "connections": [
+            {
+                "name": "hero_connection",
+                "type": "1:1",
+                "variants": [
+                    {
+                        "determinant": { "hero_type": "human" },
+                        "destination_collection": "human_collection",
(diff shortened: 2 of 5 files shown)


test/extra/connections_1_1_name_clash.test.lua
```diff
@@ -0,0 +1,367 @@
...
+local test = require('tap').test('connections 1:1 name clash')
+
+-- require in-repo version of graphql/ sources despite current working directory
+package.path = fio.abspath(debug.getinfo(1).source:match("@?(.*/)")
+    :gsub('/./', '/'):gsub('/+$', '')) .. '/../../?.lua' .. ';' ..
+    package.path
+
+local graphql = require('graphql')
+
+box.cfg{ wal_mode = "none" }
```
Totktonada: Why?

graphql/convert_schema/types.lua
```diff
@@ -373,7 +373,7 @@ function types.convert(state, avro_schema, opts)
     if context == nil then
         context = {
             field_name = nil,
-            path = {},
+            path = {'results'}
```
Totktonada: By the way, I prefer to have dangling commas, it makes further diffs easier to read.

test/extra/connections_1_1_name_clash.test.lua
```diff
@@ -0,0 +1,367 @@
...
+            service_fields = {},
+            fields = { 'hero_id' },
+            index_type = 'tree',
+            unique = true,
+            primary = true
+        }
+    }
+}
+
+local accessor_1 = graphql.accessor_space.new({
```
Totktonada: Proposed to move to `graphql.new`.

test/extra/connections_1_1_name_clash.test.lua
```diff
@@ -0,0 +1,367 @@
...
+        "schema_name": "starship",
+        "connections": []
+    },
+    "hero_info_collection": {
+        "schema_name": "hero_info",
+        "connections": []
+    }
+}]])
+
+local accessor_2 = graphql.accessor_space.new({
```
Totktonada: Proposed to move to `graphql.new`.

test/extra/connections_1_1_name_clash.test.lua
```diff
@@ -0,0 +1,367 @@
...
+
+local gql_wrapper_2 = graphql.new({
+    schemas = schemas,
+    collections = collections_2,
+    accessor = accessor_2,
+})
+
+test:isnt(gql_wrapper_2, nil)
+
+-- Potential lash between second ("hero_meta_collection") and third
```
Totktonada: lash → clash

test/extra/connections_1_1_name_clash.test.lua
```diff
@@ -0,0 +1,367 @@
...
+        "schema_name": "hero_info",
+        "connections": []
+    },
+    "hero_meta_collection": {
+        "schema_name": "hero_meta",
+        "connections": []
+    }
+}]])
+
+local accessor_3 = graphql.accessor_space.new({
```
Totktonada: Proposed to move to `graphql.new`.

Totktonada (approved):
LGTM.

Don’t forget to update instrospection.test.lua before merge.

Minor comments are below.

graphql/convert_schema/types.lua
SudoBobo: Ok, reasonable, fixed as you say.

test/extra/connections_1_1_name_clash.test.lua
SudoBobo: I put it because it was in all other /extra tests.
As original author of this tests says, `wal_mode = "mode"` speed ups tests. But in this test it is useless.
Removed.

test/extra/connections_1_1_name_clash.test.lua
SudoBobo: Ok, fixed.

test/extra/connections_1_1_name_clash.test.lua
SudoBobo: Fxd.

test/extra/connections_1_1_name_clash.test.lua
SudoBobo: Fxd.

test/extra/connections_1_1_name_clash.test.lua
SudoBobo: Fxd.

test/extra/connections_1_1_name_clash.test.lua
Totktonada: Ouch, I didn’t know that this is common way to write extra tests. So, I vote to get it back for consistency.

SudoBobo: Ok then.

PR #209, opened 2018-09-03
graphql: implement c-style expressions parser

Hollow111:
Needed for #13

### DIFF
--- a/graphql/expressions.lua
+++ b/graphql/expressions.lua
@@ -1,0 +1,233 @@
+local lpeg = require('lulpeg')
+
+local expressions = {}
+local P, R, S, V = lpeg.P, lpeg.R, lpeg.S, lpeg.V
+local C = lpeg.C
+
+-- Some special symbols.
+local space_symbol = S(' \t\r\n')
+local eof = P(-1)
+local spaces = space_symbol ^ 0
+
+-- Possible identifier patterns:
+--   1) Number.
+local digit = R('09')
+local integer = R('19') * digit ^ 0
+local decimal = (digit ^ 1) * '.' * (digit ^ 1)
+local number = decimal + integer
+--   2) Boolean.
+local bool = P('false') + P('true')
+--   3) String.
+local string = P('"') * C((P('\\"') + 1 - P('"')) ^ 0) * P('"')
+--   4) Variable.
+local identifier = ('_' + R('az', 'AZ')) * ('_' + R('09', 'az', 'AZ')) ^ 0
+local variable_name = identifier
+--   5) Object's field path.
+local field_path = identifier * ('.' * identifier) ^ 0
+
+-- Possible logical function patterns:
+--   1) is_null.
+local is_null = P('is_null')
+--   2) not_null.
+local not_null = P('not_null')
+--   3) regexp.
+local regexp = P('regexp')
+
+-- Possible unary operator patterns:
+--   1) Logical negation.
+local negation = P('!')
+--   2) Unary minus.
+local unary_minus = P('-')
+--   3) Unary plus.
+local unary_plus = P('+')
+
+-- Possible binary operator patterns:
+--   1) Logical and.
+local logic_and = P('&&')
+--   2) logical or.
+local logic_or = P('||')
+--   3) +
+local addition = P('+')
+--   4) -
+local subtraction = P('-')
+--   5) ==
+local eq = P('==')
+--   6) !=
+local not_eq = P('!=')
+--   7) >
+local gt = P('>')
+--   8) >=
+local ge = P('>=')
+--   9) <
+local lt = P('<')
+--   10) <=
+local le = P('<=')
+
+--- Utility functions:
+---     1) Vararg iterator.
+
+
+local function vararg(...)
+    local i = 0
+    local t = {}
+    local limit
+    local function iter(...)
+        i = i + 1
+        if i > limit then return end
+        return i, t[i]
+    end
+
+    i = 0
+    limit = select("#", ...)
+    for n = 1, limit do
+        t[n] = select(n, ...)
+    end
+    for n = limit + 1, #t do
+        t[n] = nil
+    end
+    return iter
+end
+
+-- AST nodes generating functions.
+local function identical(arg)
+    return arg
+end
+
+local identical_node = identical
+
+local op_name = identical
+
+local function root_expr_node(expr)
+    return {
+        kind = 'root_expression',
+        expr = expr
+    }
+end
+
+local function bin_op_node(...)
+    if select('#', ...) == 1 then
+        return select(1, ...)
+    end
+    local operators = {}
+    local operands = {}
+    for i, v in vararg(...) do
+        if i % 2 == 0 then
+            table.insert(operators, v)
+        else
+            table.insert(operands, v)
+        end
+    end
+    return {
+        kind = 'binary_operations',
+        operators = operators,
+        operands = operands
+    }
+end
+
+local function unary_op_node(unary_operator, operand_1)
+    return {
+        kind = 'unary_operation',
+        op = unary_operator,
+        node = operand_1
+    }
+end
+
+local function func_node(name, ...)
+    local args = {...}
+    return {
+        kind = 'func',
+        name = name,
+        args = args
+    }
+end
+
+local function number_node(value)
+    return {
+        kind = 'const',
+        value_class = 'number',
+        value = value
+    }
+end
+
+local function string_node(value)
+    return {
+        kind = 'const',
+        value_class = 'string',
+        value = value
+    }
+end
+
+local function bool_node(value)
+    return {
+        kind = 'const',
+        value_class = 'bool',
+        value = value
+    }
+end
+
+local function variable_node(name)
+    return {
+        kind = 'variable',
+        name = name
+    }
+end
+
+local function path_node(path)
+    return {
+        kind = 'object_field',
+        path = path
+    }
+end
+
+-- Patterns returning corresponding nodes (note that all of them
+-- start with '_').
+local _number = number / number_node
+local _bool = bool / bool_node
+local _string = string / string_node
+local _variable = '$' * C(variable_name) / variable_node
+local _field_path = field_path / path_node
+local _literal = _bool + _number + _string
+
+local _logic_or = logic_or / op_name
+local _logic_and = logic_and / op_name
+local _comparison_op = (eq + not_eq + ge + gt + le + lt) / op_name
+local _arithmetic_op = (addition + subtraction) / op_name
+local _unary_op = (negation + unary_minus + unary_plus) / op_name
+local _functions = (is_null + not_null + regexp) / identical
+
+-- Grammar rules for C-style expressions positioned ascending in
+-- terms of priority.
+local expression_grammar = P {
+    'init_expr',
+    init_expr = V('expr') * eof / root_expr_node,
+    expr = spaces * V('log_expr_or') * spaces / identical_node,
+
+    log_expr_or = V('log_expr_and') * (spaces * _logic_or *
+                  spaces * V('log_expr_and')) ^ 0 / bin_op_node,
+    log_expr_and = V('comparison') * (spaces * _logic_and * spaces *
+                   V('comparison')) ^ 0 / bin_op_node,
+    comparison = V('arithmetic_expr') * (spaces * _comparison_op * spaces *
+                 V('arithmetic_expr')) ^ 0 / bin_op_node,
+    arithmetic_expr = V('unary_expr') * (spaces * _arithmetic_op * spaces *
+                      V('unary_expr')) ^ 0 / bin_op_node,
+
+    unary_expr = (_unary_op * V('first_prio') / unary_op_node) +
+                 (V('first_prio') / identical_node),
+    first_prio = (V('func') + V('value_terminal') + '(' * spaces * V('expr') *
+                  spaces * ')') / identical_node,
+    func = _functions * '(' * spaces * V('value_terminal') * (spaces * ',' *
+           spaces * V('value_terminal')) ^ 0 * spaces * ')' / func_node,
+    value_terminal = (_literal + _variable + _field_path) / identical_node
+}
+
+--- Parse given string which supposed to be a c-style expression.
+---
+--- @tparam str string representation of expression.
+---
+--- @treturn syntax tree.
+function expressions.parse(str)
+    assert(type(str) == 'string', 'parser expects a string')
+    return expression_grammar:match(str) or error('syntax error')
+end
+
+return expressions
--- a/graphql/expressions.lua
+++ b/graphql/expressions.lua
@@ -1,4 +1,16 @@
 local lpeg = require('lulpeg')
+local utils = require('graphql.utils')
+
+--- NOTE: Functions that worth moving out into other modules:
+---       1) Regexp implementation (also lies inside of
+---          accessor_general.lua)
+---       2) Vararg iterator.
+
+--- TODO:
+---     1) Validation.
+---     2) Big numbers.
+---     3) Absence of variable in context.variables is equal
+---        to the situation when context.variables.var == nil.
 
 local expressions = {}
 local P, R, S, V = lpeg.P, lpeg.R, lpeg.S, lpeg.V
@@ -28,8 +40,8 @@ local field_path = identifier * ('.' * identifier) ^ 0
 -- Possible logical function patterns:
 --   1) is_null.
 local is_null = P('is_null')
---   2) not_null.
-local not_null = P('not_null')
+--   2) is_not_null.
+local is_not_null = P('is_not_null')
 --   3) regexp.
 local regexp = P('regexp')
 
@@ -63,31 +75,6 @@ local lt = P('<')
 --   10) <=
 local le = P('<=')
 
---- Utility functions:
----     1) Vararg iterator.
-
-
-local function vararg(...)
-    local i = 0
-    local t = {}
-    local limit
-    local function iter(...)
-        i = i + 1
-        if i > limit then return end
-        return i, t[i]
-    end
-
-    i = 0
-    limit = select("#", ...)
-    for n = 1, limit do
-        t[n] = select(n, ...)
-    end
-    for n = limit + 1, #t do
-        t[n] = nil
-    end
-    return iter
-end
-
 -- AST nodes generating functions.
 local function identical(arg)
     return arg
@@ -110,7 +97,8 @@ local function bin_op_node(...)
     end
     local operators = {}
     local operands = {}
-    for i, v in vararg(...) do
+    for i = 1, select('#', ...) do
+        local v = select(i, ...)
         if i % 2 == 0 then
             table.insert(operators, v)
         else
@@ -193,7 +181,7 @@ local _logic_and = logic_and / op_name
 local _comparison_op = (eq + not_eq + ge + gt + le + lt) / op_name
 local _arithmetic_op = (addition + subtraction) / op_name
 local _unary_op = (negation + unary_minus + unary_plus) / op_name
-local _functions = (is_null + not_null + regexp) / identical
+local _functions = (is_null + is_not_null + regexp) / identical
 
 -- Grammar rules for C-style expressions positioned ascending in
 -- terms of priority.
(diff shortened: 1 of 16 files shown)


graphql/core/init.lua
```diff
@@ -7,5 +7,6 @@ graphql.types = require(path .. '.types')
 graphql.schema = require(path .. '.schema')
 graphql.validate = require(path .. '.validate')
 graphql.execute = require(path .. '.execute')
+graphql.expressions = require(path .. '.expressions')
```
Totktonada: Expressions module should not be part of 'core' module.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
+local lpeg = require('lulpeg')
+local P, R, S, V = lpeg.P, lpeg.R, lpeg.S, lpeg.V
+local C = lpeg.C
+
+-- Some special symbols.
+local space_symbol = S(' \t\r\n')
+local eof = P(-1)
+local some_spaces = space_symbol ^ 0
```
Totktonada: Nitpicking: words like 'some' does not add any sense to a name, so it just increases a length of an identifier and decreases readability of the code.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
+local lpeg = require('lulpeg')
+local P, R, S, V = lpeg.P, lpeg.R, lpeg.S, lpeg.V
+local C = lpeg.C
+
+-- Some special symbols.
+local space_symbol = S(' \t\r\n')
```
Totktonada, line 15: Nitpicking: why not just 'space' or 'whitespace'?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+local eof = P(-1)
+local some_spaces = space_symbol ^ 0
+
+--
+-- Possible identifier patterns:
+--   1) Number.
+local digit = R('09')
+local integer = (S('-+') ^ -1) * (digit ^ 1)
```
Totktonada: 1. Unary plus / minus before a constant can be handled with unary operations. Are parsing it here does add any additional value? It seems for me as premature optimization.
2. Should we consider decimal integers as only ones that are start from [1-9]? Just in case we'll add support 0\d+ form for octet numbers in the future.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+local decimal = integer * '.' * (digit ^ 1)
+local number_pat = decimal + integer
```
Totktonada: It seems you are use '_pat' suffix inconsistently: not all patterns have it in the name. Maybe this suffix is redundant at all?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+--   2) Boolean.
+local bool_pat = P('false') + P('true')
+--   3) String.
+local string_pat = P'\"' * C((P'\\"' + 1 - S'\"') ^ 0) * P'"'
```
Totktonada: 1. `'\"'` is the same as `'"'` in lua (no need to escape a regular symbol).
2. Nit: you are use P('str') and P'both'. Prefer one form consistently: suggested to use P('str'), because it matches our lua code style.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+--   4) Variable.
+local variable_name = ('_' + R('az', 'AZ')) *
+                      (('_' + R('09', 'az', 'AZ')) ^ 0)
+--   5) Object's field path.
+local field = ('_' + R('az', 'AZ')) * ('_' + R('09', 'az', 'AZ')) ^ 0
```
Totktonada: Nit: we can define 'identifier' pattern once and reuse it for variables and fields both.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+local field_path = field * ('.' * field) ^ 0
+
+--
+-- Possible logical function patterns:
+--   1) is_null.
+local is_null = P('is_null')
```
Totktonada: Nit: can be S instead of P?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+--   2) not_null.
+local not_null = P('not_null')
+--   3) regexp.
+local regexp = P('regexp')
+
+--
+-- Possible unary operator patterns:
+--   1) Logical negation.
+local negation = P('!')
```
Totktonada: Unary minus / plus?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+
+-- Possible binary operator patterns:
+--   1) Logical and.
+local logic_and = P('&&')
```
Totktonada: Nit: can be S instead of P?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+-- AST nodes generating functions.
+--
+local function identical(arg)
+    return arg
+end
+
+local function root_expr_node(expr)
+    return {
+        kind = 'root_expression',
+        entity = expr
```
Totktonada: Nit: I would be prefer 'value' over 'entity' if choose across general words, but maybe speific 'expr' is better (because it is specific).

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+local ge = P('>=')
+--   9) <
+local lt = P('<')
+--   10) <=
+local le = P('<=')
+
+--
+-- AST nodes generating functions.
+--
+local function identical(arg)
```
Totktonada: Inconsistency: '_node' suffix is not used here.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+        entity = expr
+    }
+end
+
+local function bin_op_node(...)
+    local args = {...}
+    if select('#', ...) == 1 then
+        return identical(args[1])
+    end
+    local operators, nodes = {}, {}
```
Totktonada: Nit: we use multiple assignment across the code, but not multiple declaration. I think it would be more consistent to use two separate declarations inside ths codebase.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+    for i, v in ipairs(args) do
```
Totktonada: Nit: we can avoid constructing `args` table, consider 'Issue #8: Iterating over Varargs' in [1].

[1]: http://lua-users.org/wiki/VarargTheSecondClassCitizen

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+local function func_node(name, ...)
+    local args = {...}
+    return {
+        kind = 'function',
+        name = name,
+        args = args
+    }
+end
+
+local function op_name(string)
```
Totktonada: Nit: `op_name = identical` is better?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+    return string
+end
+
+local function number_node(value)
+    return {
+        kind = 'const',
+        data_class = 'number',
```
Totktonada: Nit: `value_class` or just `class`?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+--
+-- Grammar rules for C-style expressions positioned ascending in
+-- terms of priority.
+--
+local expression_grammar = P {
+    'init_expr',
+    init_expr = V('expr') * eof / root_expr_node,
+    expr = some_spaces * V('log_expr_or') * some_spaces / identical,
+
+    log_expr_or = V('log_expr_and') * (some_spaces * _logic_or *
```
Totktonada: Can you please give me your example of a grammar rule variant which is like A = B | A. I want to understand the problem with it and maybe do some experimenting.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+    comparison = V('arithmetic_expr') * (some_spaces * _comparison_op *
+                 some_spaces * V('arithmetic_expr')) ^ 0 / bin_op_node,
+    arithmetic_expr = V('unary_expr') * (some_spaces * _arithmetic_op *
+                      some_spaces * V('unary_expr')) ^ 0 / bin_op_node,
+
+    unary_expr = (_unary_op * V('first_prio') / unary_op_node) +
+            (V('first_prio') / identical),
+    first_prio = (V('_function') + V('identifier') + '(' * some_spaces *
+            V('expr') * some_spaces * ')') / identical,
+    _function = _functions * '(' * some_spaces * (V('identifier') *
```
Totktonada: Nit: it seems you are use B * (B+C)^0 form mostly, but here (B+C)^0 * B is used. Are that intentional?

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+            some_spaces * ',' * some_spaces) ^ 0 * V('identifier')
+            * some_spaces * ')' / func_node,
+    identifier = (_bool + _number + _string + _variable + _field_path) /
```
Totktonada: Nit: bool, number and string are constants, but not identifiers. Variable and field are identifiers. A field path is not identifier, strictly speaking. It is just naming question, but strict terminology is our friend when we investgate a code base that are not known for us.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+--
+--
+
+
+--
+-- Parsing function.
+--
+return function(str)
+    assert(type(str) == 'string', 'parser expects a string')
+    return expression_grammar:match(str) or error('syntax error', 2)
```
Totktonada: 1. Are lulpeg can provide extra information about non-matching rule or so? If so we should wrap it into our error to give a user as much info as we can.
2. I think we should not skip this stack frame in the traceback and should remove the second parameter of `error`.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
+#!/usr/bin/env tarantool
+
+local parse_expr = require('graphql.expressions')
```
Totktonada: Nit: personally I prefer use module name as name of its variable.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+local tap = require('tap')
+local test = tap.test('expr_tree')
+
+--
+-- Some of possible identifiers for tests. In case you wanna add
```
Totktonada: Nit: I would avoid such spoken language words as 'wanna' in the code.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+-- any new identifiers for testing just add them here to the
+-- corresponding tables.
+--
+local identifiers = {
+    {'const','false', 'bool'},
```
Totktonada: Nit: missed whitespace before the first comma.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+    {'const', 'true', 'bool'},
+    {'const', '11', 'number'},
+    {'const', '22.8', 'number'},
+    {'const','something', 'string'},
+    {'variable', '$variable'},
+    {'field_path', 'some_path.to.this.field'}
```
Totktonada: Nit: I prefer to add trailing comma in lists in pep-line formatting, lua supports that. It produces a more clear diff in case for a future change.

graphql/expressions.lua
```diff
@@ -0,0 +1,217 @@
...
+
+--
+--
+--
+
+
+--
+-- Parsing function.
+--
+return function(str)
```
Totktonada: Nit: please, return a table as the module value, it is how we make our modules over all projects. 'graphql.core' is the exception, because its origin is graphql-lua project with a different code style.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+    for i_2, id_2 in ipairs(identifiers) do
+        for _, op in ipairs(binary_operators) do
+            local first_identifier, second_identifier = id_1[2], id_2[2]
+            if id_1[3] == 'string' then
+                first_identifier = '\"' .. first_identifier .. '\"'
+            end
+            if id_2[3] == 'string' then
+                second_identifier = '\"' .. second_identifier .. '\"'
+            end
+            local bin_operation = first_identifier .. ' '.. op .. ' ' ..
```
Totktonada: Nit: ('%s %s %s'):format(...) looks better. BTW, it is part of our lua code style.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+
+--
+-- Bunch of tests for all the sets of binary operator and its
+-- operands.
+--
+for i_1, id_1 in ipairs(identifiers) do
+    for i_2, id_2 in ipairs(identifiers) do
+        for _, op in ipairs(binary_operators) do
+            local first_identifier, second_identifier = id_1[2], id_2[2]
+            if id_1[3] == 'string' then
```
Totktonada: Nit: use `local ID_TYPE = 3`.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+test:plan(1046)
+
+--
+-- Bunch of tests for all the sets of binary operator and its
+-- operands.
+--
+for i_1, id_1 in ipairs(identifiers) do
+    for i_2, id_2 in ipairs(identifiers) do
+        for _, op in ipairs(binary_operators) do
+            local first_identifier, second_identifier = id_1[2], id_2[2]
```
Totktonada: Nit: split to the two lines.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+                args = {arg_node_1, arg_node_2}
+            }
+        }
+        test:is_deeply(ast, expected, 'ast_func_regexp' .. tostring(i_1) ..
+                       tostring(i_2))
+    end
+end
+
+local ast = parse_expr('true && ($variable + 7 <= field.path_1 || ' ..
+                       '!(\"abc\" > \"abd\")) && !false')
```
Totktonada: Nit: escaping of double quotes is redundant.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+                op = '!',
+                node = {kind = 'const', data_class = 'bool', value = 'false'}
+            }
+        }
+    }
+}
+test:is_deeply(ast, expected, 'ast_handwritten_test_1')
+
+
+-- Tested if the same prioirty binary operators in the "same
```
Totktonada: Typo: prioirty → priority.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+-- amount of brackets" are actually different nodes and if the
```
Totktonada: if the..?

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+                }
+            }
+        }
+    }
+}
+
+test:is_deeply(ast, expected, 'ast_handwritten_test_3')
+
+--require('yaml').cfg{encode_use_tostring = true}
+--print(require('yaml').encode(parse_expr("false || false")))
```
Totktonada: No newline at the end: please, configure your text editor to produce it. A text file in unix is a sequence of lines and a line is the symbols sequence ends with the newline symbol.

test/unit/suite.ini
```diff
@@ -0,0 +1,4 @@
+[default]
+core = app
+description = tests on c-style expressions
+is_parallel = True
```
Totktonada: No newline at the end.

test/unit/expr_tree.test.lua
```diff
@@ -0,0 +1,396 @@
...
+    '-',
+    '==',
+    '!=',
+    '>',
+    '>=',
+    '<',
+    '<='
+}
+
+test:plan(1046)
```
Totktonada: You are rock!

Totktonada (commented):
Thanks for the patch!

I think it has quite good quality already, but left thoughts on possible improvements (some of them are just for consistency with the rest code base) and rice some questions. BTW, questions from a reviewer often are good candidates to describe things in a comment or a commit message.

Use branch inside the repository instead of a fork: it eases review of the changes locally (just git checkout).

It is good practice to add the prefix 'WIP: ' for pull requests in work to avoid misunderstanding of the status and prevent premature check-in.

Please, consider comments below. Consider they as recommendations that are debatable. Consider a test comments with lower priority then ones of the code.

graphql/expressions.lua
Hollow111: S stands for 'Set', which matches any character from the given argument string.

graphql/expressions.lua
Hollow111: Because `whitespace` or a `space` is a `' '`.

graphql/expressions.lua
Hollow111: Changed to 'spaces'.

graphql/expressions.lua
Hollow111: You're right thank you for remark. Removed this suffix.

graphql/expressions.lua
Hollow111: 1. Changed it, did not really know it, thank you.
2. I did not notice that. Changed to P('str') form.
3. Changed `S'\"'` to `P('"')`.

graphql/expressions.lua
Hollow111: I've added `identical_node` function (just a copy of `identical`) that is now used for passing nodes. Though as you can notice `identical` is not only used on this purpose, hence there are still some usages of it.

graphql/expressions.lua
Hollow111: Yes, I guess 'expr' suits better.

graphql/expressions.lua
Hollow111: Changed.

graphql/expressions.lua
Hollow111: Thank you for the advice.
1. Changed `return identical(args[1])` to `return select(1, ...)`.
2. Took iterator from there and used it.

graphql/expressions.lua
Hollow111: Changed it to `value_class`.

graphql/expressions.lua
Hollow111: If I'm not wrong it was:
```
log_expr_or = (V('log_expr_and') * spaces * _logic_or * spaces *
                  V('log_expr_or') / bin_op_node) + (V('log_expr_and') /
                  identical)
```
function `bin_op_node` was looking like this:
```
local function bin_op_node(left_operand, operator, right_operand)
    return {
        kind = 'binary_operation',
        op = operator,
        node_1 = left_operand,
        node_2 = right_operand,
    }
end
```

graphql/expressions.lua
Hollow111: It seems you do not fully understand the LuLPeg grammar: B * (B + C) ^ 0 is not the same as (B + C) ^ 0 * B.
`*` is used to "concatenate" patterns in corresponding order.

graphql/expressions.lua
Hollow111: 1. Probably I'm wrong but i can't find anything about it in docs or somewhere else.
2. Fixed it.

test/unit/expr_tree.test.lua
Hollow111: Not even close!

test/unit/expr_tree.test.lua
Hollow111: Nevermind.

graphql/expressions.lua
Hollow111: 1. Changed `identifier` rule to `value_terminal`.
2. United `_bool + _number + _string` into `_literal`

graphql/expressions.lua
Hollow111: 1. No it doesn't add any additional value. I didn't really consider them as operators just added 'positive' and 'negative' number values. Changed to an unary operation & I guess it will be possible to use them for identifiers containing 'number value'.
2. Why not? Fixed according to the idea.

graphql/expressions.lua
Totktonada: Why do you think so? There are many different space symbols, which are especially useful in typography: http://jkorpela.fi/chars/spaces.html

graphql/expressions.lua
Totktonada: Ouch, it was just typos, sorry. I mean left/right associalivity. B * (OP * B)^0 vs (B * OP)^0 * B.

Hollow111: They're still different. First means "match pattern B which is followed by at least 0 repetitions of pattern OP * B" whilst second means "match at least 0 repetitions of pattern OP * B followed by pattern B".

graphql/expressions.lua
Hollow111: I do not say that there are the only 'space' symbol. I mean that `\t` for example isn't a 'whitespace' or a 'space', it's a 'space symbol'.

Totktonada: Ok.

graphql/expressions.lua
Totktonada: Yep, except the second 'OP * B' in your phrase should be 'B * OP'. The question is why do you use this kind of a rule here, but other in binary operations?

Hollow111: Fixed.

test/unit/expr_execute.test.lua
```diff
@@ -0,0 +1,36 @@
...
+str = 'true && ($variable + 7 == field.path_1 && !(2399>23941)) && !false'
+expr = expressions.new(str, {node_1 = {node_2 = 21},
+    field = {path_1 = 28}}, {variable = 21})
+test:is(expr.execute, true)
+
+test:is(expressions.compile_and_execute(str, {node_1 = {node_2 = 21},
+    field = {path_1 = 28}},
+    {variable = 21}), true)
+
+str = 'false || (regexp("abc", $var)) && (is_null(path.ddf))|| not_null(path.ddf)'
```
Totktonada: I think we should use `is` consistently: is_null and is_not_null.

graphql/expressions.lua
```diff
@@ -1,4 +1,20 @@
 local lpeg = require('lulpeg')
+local utils = require('graphql.utils')
+local rex, is_pcre2 = utils.optional_require_rex()
+local bit = require('bit')
+
+--- NOTE: Functions that worth moving out into other modules:
+---       1) Regexp implementation (also lies inside of
+---          accessor_general.lua)
+---       2) Vararg iterator.
```
Totktonada: Maybe `fun` library contains something like this function.

graphql/expressions.lua
```diff
@@ -1,4 +1,20 @@
...
+
+--- TODO:
+---     1) Validation.
+---     2) Big numbers.
+---     3) Absence of variable in context.variables is equal
+---        to the situation when context.variables.var == nil.
+---     4) In case you change string inside of an expression
+---        object you need to manually change ast in object.
```
Totktonada: Don't get this point.

graphql/expressions.lua
```diff
@@ -230,4 +266,207 @@ function expressions.parse(str)
...
 end
 
+--function expressions.validate(context)
+--
+--end
+
+--- Recursively execute the syntax subtree. Of course it can be
+--- syntax tree itself.
+---
+--- @tparam node    node to be executed.
```
Totktonada: Please, read ldoc documentation. `tparam` assumes type, then name.

graphql/expressions.lua
```diff
@@ -230,4 +266,207 @@ function expressions.parse(str)
...
+--- @tparam variables list of variables.
+---
+--- @treturn expression object.
+function expressions.new(str, object, variables)
+    return setmetatable({
+        string = str,
+        object = object,
+        variables = variables,
+    }, {
+        __index = function(self, key)
```
Totktonada: Please, rewrite with a table instead of the function. It si for consistency with other code and to get proper error in case of trying to call a non-existent function.

graphql/expressions.lua
```diff
@@ -230,4 +266,207 @@ function expressions.parse(str)
...
+end
+
+--- Create a new c-style expression object.
+---
+--- @tparam str       string representation of expression.
+--- @tparam object    object considered inside of an expression.
+--- @tparam variables list of variables.
+---
+--- @treturn expression object.
+function expressions.new(str, object, variables)
```
Totktonada: Proposed API:

expressions:
- new(<string>) -> compiled_expr
- execute(<string>[, object[, variables]]) -> result // see note 1

compiled_expr:
- ast // data, not function
- execute(self[, object[, variables]]) -> result

Note 1: `compile_and_execute` local function should be exposed as `expressions.execute`. It will match box.sql.execute and graphql.execute approach.

graphql/expressions.lua
```diff
@@ -230,4 +266,207 @@ function expressions.parse(str)
...
+---
+--- @treturn expression value.
+function expressions.compile_and_execute(str, object, variables)
+    local context = expressions.new(str, object, variables)
+    return expressions.execute(context.ast, context)
+end
+
+--- Create a new c-style expression object.
+---
+--- @tparam str       string representation of expression.
```
Totktonada: [t]param / [t]return expected to have no a period at the end, afair.

graphql/expressions.lua
```diff
@@ -220,6 +236,26 @@ local expression_grammar = P {
...
 }
 
+-- Add one number to another.
+--
+-- It may be changed after introduction of "big ints".
+--
+-- @param operand_1
+-- @param operand_2
+--
+local function sum(operand_1, operand_2)
```
Totktonada: What is purpose of these wrappers?

graphql/expressions.lua
```diff
@@ -230,4 +266,207 @@ function expressions.parse(str)
...
+            if op == '&&' then
+                prev = prev and second_operand
+            end
+
+            -- Logical or.
+            if op == '||' then
+                prev = prev or second_operand
+            end
+
+            -- Equal.
```
Totktonada, line 342: TBD: think how to handle nil / box.NULL with equal and other operation. Like SQL does, ternary logic?

test/unit/expr_execute.test.lua
```diff
@@ -0,0 +1,36 @@
...
+
+test:is(expressions.compile_and_execute(str, {path = {ddf = 'hi_there'}},
+    {var = 'abc'}), true)
+
+str = 'false || (regexp("abc", $var)) && (is_null(path.ddf))|| is_null(path.ddf)'
+expr = expressions.new(str, {path = {ddf = 'hi_there'}}, {var = 'abc'})
+test:is(expr.execute, false)
+
+test:is(expressions.compile_and_execute(str, {path = {ddf = 'hi_there'}},
+    {var = 'abc'}), false)
```
Totktonada: Don't forget test:check at the end. Also see other tests, it is better to pass it to exit code.

graphql/expressions.lua
```diff
@@ -1,4 +1,20 @@
 local lpeg = require('lulpeg')
+local utils = require('graphql.utils')
+local rex, is_pcre2 = utils.optional_require_rex()
+local bit = require('bit')
+
+--- NOTE: Functions that worth moving out into other modules:
+---       1) Regexp implementation (also lies inside of
```
Totktonada: Just move it to graphql/utils.lua and remove copies?

Totktonada (commented):
I look briefly into the executor. Please, consider comments below. The most important one is API proposal.

Why do you say here that the commit implements expression parser? It is not so, **this** commit implements only executor.

graphql/expressions.lua
Totktonada: 1. Please, test it.

graphql/expressions.lua
Totktonada: Missed the comment. Ok.

Totktonada:
@Hollow111 BTW, how to break a line within an expression?

graphql/expressions.lua
Hollow111: I mean that it's possible to change `expr.raw` manually, which turns out to have old `expr.ast`. I'm not really sured if we should be able to handle this type of situations.

Hollow111:
> @Hollow111 BTW, how to break a line within an expression?

If i get you right: just use multiline lua-string and break lines as usual (it behaves the same as space).

e.g.:

```
[[
    $var1 &&
    path.to.some.node
]]
```

graphql/expressions.lua
Totktonada: Direct changing of an object data is definitely inappropriate use of the module. We can protect a user from this kind of action (using __newindex field of a metatable), but I don't think we should: it would introduce unnecessary complexity to the code.

Totktonada:
@Hollow111 Thanks! It seems that our graphql parser is not ready for this. I have filed the issue about is: #223. It should not be part of this PR, though.

graphql/expressions.lua
Hollow111: Ok then.

graphql/expressions.lua
Hollow111: Can't find something to change for.

Totktonada: I look into the usage. You are need just iterate over arguments of a function with variable number of arguments. So just do that:

```
for n=1,select('#',...) do
  local e = select(n,...)
end
```

It is from 'Issue #8: Iterating over Varargs' section of [1].

[1]: http://lua-users.org/wiki/VarargTheSecondClassCitizen

This will simplify the code, so I vote for that.

PR #216, opened 2018-09-07
Add graphql subscription

filonenko-mikhail:
Subscription provide pubsub mechanism for using
in fullduplex transports.

### DIFF
--- a/graphql/core/introspection.lua
+++ b/graphql/core/introspection.lua
@@ -29,7 +29,7 @@ __Schema = types.object({
 
   description = util.trim [[
     A GraphQL Schema defines the capabilities of a GraphQL server. It exposes all available types
-    and directives on the server, as well as the entry points for query and mutation operations.
+    and directives on the server, as well as the entry points for query, mutation and subscription operations.
   ]],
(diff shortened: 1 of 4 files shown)


Totktonada:
@filonenko-mikhail An umbrella issue re subscriptions support need to be filed. It should contain at least high-level paln want exactly we want to support. Short questionary to help you with that:

- Real / semi-real use cases illustrating gains from using subscriptions.
- How a user have to configure a channel(s) to pass events to.
    - And how she/he have to choose a channel in a subscription request (an argument with a channel name from a configuration?).
    - Should channels configuration be integrated with some of our http module facilities?
    - Should channels configuration allows to pass custom callbacks?
- What events should be handled:
    - Mutations via the current graphql instance?
    - Tarantool's box triggers?
    - Custom events provided via configuration?
    - Any others?

I'll leave the PR open and stalled until we'll elaborate the plan of the feature and the plan how the implementation should look.

filonenko-mikhail:
I rethink subscription protocol in graphql. It is over engineered so much.

PR #220, opened 2018-09-09
Kh/vshard

Khatskevich:
Add vshard accessor to gql.


graphql/accessor_general.lua
```diff
@@ -1245,8 +1245,6 @@ local function insert_internal(self, collection_name, from, filter, args, extra)
     assert(next(args) == nil, err_msg)
     assert(next(extra.extra_args, next(extra.extra_args)) == nil, err_msg)
     check(from, 'from', 'table')
```
Totktonada, line 1260: The check that `from` contains the same values as the object to insert is needed.

Totktonada:

* Archive the repository.

PR #254, opened 2018-10-10
Give an empty multihead connection as null

### LINKED ISSUE #253: Test null values in multihead connections
What is need to be tested:

- No object(s) got for 1:1 / 1:N multihead connection (with null key as well as with disable_dangling_check options).
- No variant in a query for 1:1 / 1:N multihead connection.

Results for all cases need to be verified against generated avro-schema and tested on both executors: bfs and dfs.

Related comment: https://github.com/tarantool/graphql/issues/244#issuecomment-427090396

Totktonada:
Fixes #253.

### DIFF
--- a/graphql/bfs_executor.lua
+++ b/graphql/bfs_executor.lua
@@ -575,6 +575,7 @@ local function filter_object(object, object_type, selections, context, opts)
                 qcontext = qcontext,
                 schema = context.schema,
                 parentObject = object_type,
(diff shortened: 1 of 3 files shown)


Totktonada:
Awaiting for Pavel or Roman to ensure this PR fixes cases they are met.

Totktonada:
Pavel says his tests are okay after the change.

PR #258, opened 2018-12-15
Use built-in merger when available

Totktonada:
It is implemented for batch requests. Regular requests uses
shard.secondary_select(), which already uses merger under hood (its own
version of the merger).


Totktonada:
Blocked until the merger will be land into tarantool (see https://github.com/tarantool/tarantool/issues/3276 ), because the API can be changed during review.

Totktonada:

* Archive the repository.

PR #261, opened 2019-01-29
Add an ability to pass the request context to the 'execute' function.

Kasen:
This context is saved in the query context and pass to custom accessor functions


Totktonada:
I took the main idea and rewrote to achieve the several benefits:

* Made renames, move code here and there :)
* Wrap user context into opts in :execute(); it seems now the time to introduce this parameter for this and any future runtime options.
* Reflect :execute() change in compile_and_execute().
* Ensure all redefineable functions are receive the user context.
* Wrote a test.

So I'll close it and supersede with PR #276. Thanks!

PR #265, opened 2019-02-02
Report code coverage to coveralls.io using luacov

### LINKED ISSUE #255: Report code coverage to coveralls.io using luacov
https://github.com/moteus/luacov-coveralls
https://github.com/keplerproject/luacov

Totktonada:
Updated test-run to the version with --luacov support.

Fixes #255.


Totktonada:
It seems I'm doing something wrong. The result is VERY flaky: I got coverage 65% and 75% on travis on the same code.

Totktonada:
Found the cause: it is incorrect to just concatenate luacov.stats.out files to generate a correct luacov.stats.out file. We need to merge it as luacov do in runner.save_stats(). Updated the PR.

Totktonada:
TBD: merge https://github.com/tarantool/test-run/pull/139 , update the submodule here.
TBD (optional): use `include` configuration option (is there command line argument?) to avoid writing unnecessary statistics.

I observed a small noise in results. It was for example in utils.regexp() `if not pattern:match(string)` branch with `return false`. I don't see any external reason for that. Maybe luacov sometimes gives some noise? Cannot reproduce it locally, observed only on travis-ci.

PR #287, opened 2019-06-03
avro-schema generation: support record inside map

### LINKED ISSUE #283: Converting of  GraphQL query to Avro schema doesn't work if a connection contains a map with records
Error:
```
/usr/share/tarantool/graphql/core/query_util.lua:68: bad argument #1 to 'ipairs' (table expected, got nil)
```

Totktonada:
This is more like the quick fix. I think the right way would be create a
module to convert a GraphQL type (w/o selections) into avro-schema.

Also a union inside a map unlikely will work, but here I have fixed only
the case with a record inside a map. I'll follow it up if there will be
a demand.

Fixes #283.

### DIFF
--- a/graphql/query_to_avro.lua
+++ b/graphql/query_to_avro.lua
@@ -32,10 +32,11 @@ local gql_scalar_to_avro_index = {
     Boolean = "boolean"
 }
 
-local function gql_scalar_to_avro(fieldType)
+local function gql_scalar_to_avro(fieldType, context)
     assert(fieldType.__type == "Scalar", "GraphQL scalar field expected")
+    check(context, 'context', 'table')
     if fieldType.subtype == "Map" then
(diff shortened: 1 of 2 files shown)


test/extra/to_avro_arrays.test.lua
```diff
@@ -71,6 +77,16 @@ fields:
...
+      - name: customer_balances
+        type:
+          type: map
+          values:
+            type: record
+            name: another_balance
+            fields:
+            - name: value
+              type: int
+            namespace: Query.user_collection
```
Totktonada, line 87: Should we add one level to the namespace when passing through a map?

Totktonada: It seems avro-schema only adds a new level for record, enum and fixed types. So we can leave a namespace as is when passing over a map.

